PR #7, opened 2020-04-14
Fixed missing definitions

mark-a-potts:
While I was working on updating the GSI to use the public release version of NCEPLIBS, I found that the definitions set at the top of the CMakeLists.txt file are not added during compile. Apparently, this doesn't affect the UFS, but without the underscore (from -DUNDERSCORE), there are undefined symbols in libbufr_*.


aerorahul:
@mark-a-potts Is this PR still relevant.  If not, please close.

PR #11, opened 2020-07-18
add unit_tests

aerorahul:
This PR adds unit tests from @jbathegit .

- It includes a bugfix, where `c_hdr` were not available under build interface.
- Adds ctests for tests from @jbathegit 
- `README` is translated into `Markdown` format.
- The test data is downloaded from `https://ftp.emc.ncep.noaa.gov/public/bufr.tar`.
- added github action to run tests on push and PR.

Testing:
- All _4 and _d tests pass.  _8 tests don't. I think that is expected.
- OUT3 tests do not compile on GNU and hence cannot be tested. Compiles and runs on Intel.

**Help wanted!**
I cannot figure out why `test_OUT_3.f` won't compile with GNU and dynamic allocation.
On Intel, the test compiles.

On Intel and GNU, `_8` tests exhibit failures.  This will need careful look at the test codes itself.


mark-a-potts:
All the tests other than the _8 ones pass for me using both GNU and Intel on Hera.

aerorahul:
I think @jbathegit did not intend to run test the _8 library, but I may be wrong.
we can wait for @jbathegit to return and give us instructions oh how he wishes to proceed.

aerorahul:
Changing this to an actual PR.

aerorahul:
> This looks good to me. Worked with no problems on my generic linux box using GNU. My only question is whether we should have the ENABLE_TESTING default to ON instead of OFF.

I was just keeping the defaults as other NCEPLIBS.  @kgerheiser has the defaults to `ENABLE_TESTS=OFF`.  A push and PR triggers tests on github.
I am happy to turn it ON, @kgerheiser?

PR #12, opened 2020-08-05
Feature/c wrapper

rmclaren:
Created C wrapper for some very basic NCEPLIBS-bufr functions. The purpose of this wrapper is to make it easy to use NCEPLIBS-bufr from a C or C++ application.


aerorahul:
> This looks fine to me, though I have some qualms about having a period in the middle of a file name. Could we change the names to bufr_interface.* ?

@mark-a-potts I am the one to be blamed for that.  Although, I will pass the blame to JEDI.  JEDI follows that convention for C/Fortran interfaces. Here is an [example](https://github.com/JCSDA/ufo/tree/develop/src/ufo).

If you really really really have strong aversion to it, we can make the change.

mark-a-potts:
Thanks, Rahul.

If it is working okay for JEDI, I am sure it will be fine here, too. It 
just set off some warning in my brain that it might cause issues of some 
sort. Probably some residual memories of working with DOS or something 
I'd rather not remember!

-M

PR #36, opened 2020-10-31
testing always on in cmake

### LINKED ISSUE #19: enable_tests should be on by default
We want users of this code to run tests, so they should be enabled by default, as is the common custom.

edwardhartnett:
Fixes #19 

In this PR we enable tests in the CMake file.

The way to not run tests is simply not to use "make test". So we don't need an option to turn off building tests.


jbathegit (approved):
The changes to have testing always on in CMake look fine.

But I'm curious why you also took out the check for supported compilers in this same pull request?  I thought you said you didn't want to mix different issues in the same pull request, so shouldn't that have been a different pull request?

edwardhartnett:
Yes, you're correct, I probably should have done that in a separate PR...

Issue #54, opened 2020-11-05
update to subroutine WRITSA, including new test case


jbathegit:
In my BUFRLIB repository in VLab, I have an update to subroutine WRITSA that I'd been planning to introduce in the next version for the WCOSS and on my website. Now that I'm working to fully merge everything into the NCEPLIBS-bufr repository on GitHub, I'd like to bring this update over to GitHub and just work from there going forward.

The update to the subroutine allows  the return of up to two BUFR
messages during the same call to the subroutine, in the rare instances where more than one BUFR message could become available during such a call.  I'll include a test for for this situation as part of a new test program once I submit a pull request for this issue.

Issue #53, opened 2020-11-05
updates to subroutines READLC and WRITLC, including new test cases


jbathegit:
In my BUFRLIB repository in VLab, I have an update to subroutines READLC and WRITLC that I'd been planning to introduce in the next version for the WCOSS and on my website. Now that I'm working to fully merge everything into the NCEPLIBS-bufr repository on GitHub, I'd like to bring these updates over to GitHub and just work from there going forward.

Subroutine READLC was modified to return a "missing" character string if
the requested mnemonic isn't found in the subset.  Previously, the subroutine would return a string of all blank characters in such cases.  Likewise, subroutine WRITLC was modified to print a warning message if the requested mnemonic isn't found in the subset definition, whereas the library would previously abort in such cases.  I'll include a test for each of these cases as part of a new test program once I submit a pull request for this issue.

Issue #52, opened 2020-11-05
update to subroutine UFBSEQ, including new test case


jbathegit:
In my BUFRLIB repository in VLab, I have an update to subroutine UFBSEQ that I'd been planning to introduce in the next version for the WCOSS and on my website. Now that I'm working to fully merge everything into the NCEPLIBS-bufr repository on GitHub, I'd like to bring this update over to GitHub and just work from there going forward.

The update to the subroutine deals with a situation where the number of available data levels exceeds the amount of user-provided array space when reading from an input file.  Previously, the software would abort in such cases without returning any data whatsoever, but with the change it will now print a diagnostic and return with the number of levels equal to the amount of array space that was provided.  I have a sample test case for this situation and will include it as part of a new test program once I submit a pull request for this issue.

Issue #33, opened 2020-10-28
Unit tests for src/copybf.f, src/copymg.f, src/copysb.f, src/cpbfdx.f, src/cpdxmm.f, src/cpymem.f, src/cpyupd.f 


edwardhartnett:
src/copybf.f, src/copymg.f, src/copysb.f, src/cpbfdx.f, src/cpdxmm.f, src/cpymem.f, src/cpyupd.f are all untested

jbathegit:
I think I can cover src/cpbfdx.f within a new test program that I'm currently working on for #51, #52, #53, and #54.

edwardhartnett:
Codes with apparently zero testing include:

1. bort and it's variants.
2. copybf
3. cwbng
4. dlloctbf
5. drstpl
6. invmrg
7. mesgbf
8. mrginv
9. nevn
10. newwin
11. nwords
12. openbt
13. trybump
14. ufbcpy
15. ufbcup
16. ufbget
17. ufbin3
18. ufbinx
19. ufborv
20. ufbtam

So if these codes are important, they need to be tested.

If they are not, they need to be removed. ;-)

Getting at least a few tests for each subprogram should be a top priority unless there are urgent bugfixes or new releases that need to get out to users...

These tests should be written in the style of intest1.F90:
* They should all be *.F90 codes.
* They should return non-zero for an error, and exit normally for success.
* Scanning of the output is not necessary, so in CMakeLists.txt these are just ordinary Fortran test programs that are compiled and run, just like intest1.
* They should be well-documented as to what they are, and what they are trying to do.

jbathegit:
With #331 we now have coverage for the following subprograms from the above list:

- 2 (copybf)
- 3 (cwbmg)
- 4 (dlloctbf)
- 7 (mesgbf)
- 14 (ufbcpy)
- 16 (ufbget)
- 18 (ufbinx)
- 20 (ufbtam)

jbathegit:
With the latest updates to #331 we now also have coverage for:
- 6 (invmrg)
- 8 (mrginv)
- 9 (nevn)
- 11 (nwords)
- 13 (trybump)
- 17 (ufbin3)
- 19 (ufbovr)

edwardhartnett:
@jbathegit here's a question about copymg.

If I run this code:

```
        open(unit = 11, file = 'testfiles/IN_2', form = 'UNFORMATTED', iostat = ios)
        if (ios .ne. 0) stop 3
[...]

PR #56, opened 2020-11-06
bring over several pending WCOSS src updates, and add new test_OUT_4 program

### LINKED ISSUE #51: update to subroutine STNDRD, including new test case
In my BUFRLIB repository in VLab, I have an update to subroutine STNDRD that I'd been planning to introduce in the next version for the WCOSS and on my website.  Now that I'm working to fully merge everything into the NCEPLIBS-bufr repository on GitHub, I'd like to bring this update over to GitHub and just work from there going forward.

The update to the subroutine fixes an internal calculation when standardizing messages which contain only one subset.  It was discovered after researching a WCOSS operations failure back in March, and it was temporarily patched there, but now I'd like to bring the patch into the GitHub baseline.  I still have the failed WCOSS test case saved off from March, and I'll include it as part of a new test program once I submit a pull request for this issue.

jbathegit:
Fixes #51, #52, #53, #54 

Part of #33, because src/cpbfdx.f is called internally by WRDXTB within new test_OUT_4 program


edwardhartnett:
Does this break backward compatibility?

Any change that breaks backward compatibility is a problem...

test/test_OUT_4.f
```diff
@@ -0,0 +1,153 @@
...
+
+            CALL HOLD4WLC ( 13, smid, 'SMID' )
+            CALL WRITSA ( 13, MXBFMG, mgbf, lmgbf )
+            IF ( nsub .eq. 1 ) THEN
+              CALL WRITLC ( 13, dummystr, 'DUMMYSTR' )
+            END IF
+           
+          END DO
+
+          CALL WRITSA ( -13, MXBFMG, mgbf, lmgbf )
```
edwardhartnett, line 140: You seem to be writing a file in testing, but  don't see where the file is opened and checked for correctness...

test/test_OUT_4.f
```diff
@@ -0,0 +1,153 @@
...
+
+          print *, '        ICBFMS'
+          print *, '        HOLD4WLC'
+          print *, '        DRFINI'
+          print *, '        WRITSA'
```
edwardhartnett, line 145: In general, tests should restrict their output to something pretty short and to the point, and understand that it's purely for fun, and should never need to be looked at to tell if the test worked. Here's some output from a netCDF test, which all use similar output. I would like to see all NCEPLIBS use a similar output style too, so let's develop it.


```
(base) ed@mikado:~/netcdf-c/nc_test4$ ./tst_types 

*** Testing netCDF-4 types...
*** testing netCDF-4 new atomic types and equality...ok.
*** testing user-defined netCDF-4 enum types and equality...ok.
*** testing vara functions with new netCDF-4 atomic types...ok.
*** testing vars functions with new netCDF-4 atomic types...ok.
*** testing var1 functions with new netCDF-4 atomic types...ok.
*** testing varm functions with new netCDF-4 atomic types...ok.
*** Tests successful!
(base) ed@mikado:~/netcdf-c/nc_test4$ 
```

So each test just gets one line of output.

If I need more output while developing and/or debugging the test, I comment out the extra print lines before finishing, so the test gives reasonable and pretty output when complete.

test/test_OUT_4.f
jbathegit: For all of these output tests, the correctness of the output is verified within test_wrapper.sh, where the ${outname}.bufr file is checked for bitwise equality vs. testfiles/${refname} using "cmp -s".  So the generated file is indeed opened and checked for correctness.

test/test_OUT_4.f
jbathegit: These print statements were put in place back before the tests were migrated to GitHub, and you're correct that at the time they were mostly used to just visually check that all of the tests passed.  They're less important now that everything is more automated in GitHub, but they still do get printed to build/Testing/LastTest.log and can serve as useful checkpoints for troubleshooting in case something goes wrong.   So I would argue to leave them in, but rest assured they aren't the primary criteria by which a test is graded as pass/fail - that is still completely automated as noted above in the previous reply ;-)

jbathegit:
Hmm OK, but then how are we supposed to explicitly test new code features without tailoring the test to those specific changes?  I thought that was the whole point of writing new test cases every time we add new code features.  And since the new test cases are only ever released with the new source code (i.e. only compatible versions of src and test are ever included together within the same delivery bundle), then there should never be any backwards-compatibility issues, unless I'm missing something here(?)

test/test_OUT_4.f
jbathegit: So just to clarify, the test output that the user sees when running "ctest" is indeed concise, reasonable and pretty as you showed in your example above.  Again, the more verbose "print *" output lines don't echo to the screen and only ever appear in the build/Testing/LastTest.log file, so the user won't ever see those unless they explicitly dig for them in the log file.

edwardhartnett:
WRT backward compatibility: when we change NCEPLIBS-bufr in such a way that existing code produces different results, that's a problem.

When we add new code that can be run to give new results, but the old code still gives the old results, we don't violate backward compatibility. In other words, are you adding functionality? Or are you changing (or removing) functionality? The former is easy, the latter requires extra work.

The first question is: are you breaking backward capability? Or are you adding new capabilities, which won't bother existing users?

So when we do a new release of NCEPLIBS-bufr, can it just be dropped in as a replacement, and give the same results?

It doesn't mean we can't break backward compatibility if we ever really need to, but it should be a last resort, and we should understand that if we do that, we create work for every user project, which will now need to check for what version of NCEPLIBS-bufr has been found, and adjust itself to either the old, or the new behavior.

jbathegit:
OK, thanks for the clarification, and in that case I think we're fine.  The code changes are to fix a couple of bugs, and to improve the behavior of the code by now returning partial output or "missing" values in cases where the code would previously abort.  So if those conditions were hit then I guess user code could now potentially behave a bit differently, but I would argue that those are still a valuable improvement, and we can just make sure everyone's aware by making sure that we always document changes in the release notes.  That way, users are kept in the loop and can decide for themselves if/when they want to migrate to the newer version or just stick with the previous version of the code that they're currently running.

That's the way we've always handled this sort of thing in the past.  The old/current version of the software is still available, so we're never forcing users to move up to the newer version of the code until they're ready, but this way it also doesn't prevent us from making improvements to the code just because there may be some users out there who can't (or don't want to bother ;-) switching to a newer version.  Again, the key differences can always be described in the release notes with any new version, and this approach has worked well for us for many years with this particular library.

edwardhartnett:
We must raise our standard a little bit when it comes to breaking backward compatibility. A mention in the RELEASE_NOTES is not enough, nor is it sufficient (though it is necessary) to have old versions available.

In practice, anyone at NOAA writing operational code using NCEPLIBS-bufr cannot stay at old versions. Nor do we want them to.

So if we break backward compatibility we are causing cost in other areas of NOAA, and that must be properly managed. (Also, it is almost always easier for us to operate in a backwards-compatible way.)

We should regard it as a requirement that output of existing user codes do not change due to an upgrade in NCEPLIBS-bufr. 

If that requirement cannot be met in some cases, we have to bring in other people to the discussion, and manage when that change is introduced, and also make sure the build systems are smart enough to handle it cleanly. So there is extra work and extra risk there. 

This library must operate like netCDF or HDF5. If you drop in a new version, your existing codes will always do the exact same thing they have always done. Users trust this, and barriers to upgrade are low. 

New functionality is controlled by new functions or new parameters in a backwards-compatible way.

test/test_OUT_4.f
edwardhartnett: OK, great.

edwardhartnett (approved):

jbathegit:
> We must raise our standard a little bit when it comes to breaking backward compatibility. A mention in the RELEASE_NOTES is not enough, nor is it sufficient (though it is necessary) to have old versions available.

I have to respectfully disagree on this point.  In my opinion, it should be sufficient to advertise any changes in RELEASE_NOTES and just let users migrate to the newer versions if/when they are ready.   This provides sort of a "carrot and stick" approach, where we don't force all users to migrate at any one particular time, but at the same time if they do end up needing a new feature(s) for something, then at that point the only way they can get it is by updating to the new version.  This approach has always worked for us in the past on the WCOSS and for supporting all of our external users.

> We should regard it as a requirement that output of existing user codes do not change due to an upgrade in NCEPLIBS-bufr.

> New functionality is controlled by new functions or new parameters in a backwards-compatible way.

I get where you're coming from with these statements, and in principle I agree with you, but please understand that NCEPLIBS-bufr is structured such that many routines are called internally at lower-levels, and sometimes never directly by users.  So if I strictly followed this rule, then that means the only way I can ever fix a bug or otherwise change certain low-level functionality in a subprogramA.f is by creating a new version (e.g. subprogramA2.f), which then in turn necessitates a new subprogramB2.f for any subprogramB.f which internally calls subprogramA.f, etc., etc., and then you end up having a bunch of copies of substantially the same code for perpetuity, which then becomes its own maintenance issue.

Believe me, I get that we need to think of users whenever we make changes, and I always do, by making any changes as benign as possible and then advertising them prominently.  And truth be told, the library as a whole is nearly 30 years old and could definitely benefit from a major refactoring effort or even a wholesale replacement (ideally using an OO language ;-), but that would be a MAJOR undertaking that should be its own separate discussion and would require resources way beyond just me.

PR #58, opened 2020-11-09
bring over last remaining WCOSS src updates, and add new test_OUT_5 program

### LINKED ISSUE #55: updates to subroutines UFDUMP and UFBDMP, including new test cases
In my BUFRLIB repository in VLab, I have an update to subroutines UFDUMP and UFBDMP that I'd been planning to introduce in the next version for the WCOSS and on my website. Now that I'm working to fully merge everything into the NCEPLIBS-bufr repository on GitHub, I'd like to bring these updates over to GitHub and just work from there going forward.

These two subroutines are used as offline diagnostic tools to generate verbose dump listings of the contents of a BUFR subset.  Subroutine UFDUMP was modified to improve the logic for sequence tracking,
and to increase the format width for integer code figure and bit number values when printing associated meaning strings.   Subroutine UFBDMP was modified to fix a bug when checking for the "missing" value in long character strings (i.e. longer than 8 bytes) and to enable printing of up to 120 characters for such strings.  I'll include a test for each of these cases as part of a new test program once I submit a pull request for this issue.

### LINKED ISSUE #57: do we still need -DENABLE_TESTS=ON?
@edwardhartnett @aerorahul 

Now that testing is enabled by default, do we still need -DENABLE_TESTS=ON in README.md (in the Installation instructions) and in .github/workflows/main.yml?  It might be cleaner if we just remove these - any thoughts?

jbathegit:
Fixes #55 
Fixes #57

These are the last remaining src updates to be migrated over to NCEPLIBS-bufr on GitHub, and they'll have no impact on any operational user codes because src/ufdump.f and src/ufbdmp.f are utilities that are only ever run off-line to generate verbose ASCII dumps of a BUFR file.  In other words, these subprograms are explicitly designed and designated for use as off-line diagnostic tools to help visualize the contents of a BUFR file.

Once this is in, I promise I'll get back to focusing on documentation! ;-)


jbathegit:
Hmm, OK looks like I had a fail in the push, even though it worked fine when I tested in my area.  Let me take a closer look.

aerorahul:
@jbathegit 
Is there a reason why `ufbdmp.f` and `ufdump.f` are not part of the library?

jbathegit:
@aerorahul  Sorry if I wasn't clear.  Those subprograms are part of the library, but the point I was trying to emphasize is that they're never called within real-time/operational codes.  They're only ever called within off-line diagnostic codes for users who want to print a verbose ASCII listing of the file.   In fact, they can actually be called interactively to print output in chunks of 10 lines at a time.

@edwardhartnett  So the issue with the automated build (which was done using Gnu) is that, apparently, when you have a line of Fortran code such as the following in src/ufdump.f:

WRITE (LUOUT, *)

which is intended to just print a blank line to the output ASCII file, it actually prints just an end-of-line character and nothing more, whereas all of my test builds (which were done using Intel) instead write a single blank space and then an end-of-line character.  So basically a difference of one blank space, and I had to use ":set list" within vi to actually see what was going on here!  So if we'd been using "diff -w" then it would have been fine, but of course "cmp -s" flags this difference.  At any rate, I think I can fix this fairly easily in a way so that it generates consistent output all compilers, and sorry about this!

aerorahul:
> @aerorahul Sorry if I wasn't clear. Those subprograms are part of the library, but the point I was trying to emphasize is that they're never called within real-time/operational codes. They're only ever called within off-line diagnostic codes for users who want to print a verbose ASCII listing of the file. In fact, they can actually be called interactively to print output in chunks of 10 lines at a time.
> 
> @edwardhartnett So the issue with the automated build (which was done using Gnu) is that, apparently, when you have a line of Fortran code such as the following in src/ufdump.f:
> 
> WRITE (LUOUT, *)
> 
> which is intended to just print a blank line to the output ASCII file, it actually prints just an end-of-line character and nothing more, whereas all of my test builds (which were done using Intel) instead write a single blank space and then an end-of-line character. So basically a difference of one blank space, and I had to use ":set list" within vi to actually see what was going on here! So if we'd been using "diff -w" then it would have been fine, but of course "cmp -s" flags this difference. At any rate, I think I can fix this fairly easily in a way so that it generates consistent output all compilers, and sorry about this!

`write(LUOUT, *) ' '` ?

jbathegit:
Actually, the first thing I'm going to try is using a "/" in the format to accomplish the same thing.  But yes that would probably also work, so I'll try that if my first idea doesn't pan out.

jbathegit:
OK done and fixed, now with a 2nd small add-on commit.  Please review whenever you have an opportunity, and sorry again for the  delay!

jbathegit:
Friendly reminder to @edwardhartnett - could you please review this pull request when you get a chance?  Or if you want I can get someone else to look it over - just let me know (and thanks!)

src/icbfms.f
```diff
@@ -27,7 +27,7 @@
 C>
 C> REMARKS:
 C>    THIS ROUTINE CALLS:        IUPM
-C>    THIS ROUTINE IS CALLED BY: RDCMPS   RDTREE   UFDUMP
+C>    THIS ROUTINE IS CALLED BY: RDCMPS   RDTREE   UFBDMP   UFDUMP
```
edwardhartnett, line 30: Can you change all function names in documentation to lower-case and with parens?

So:

`THIS ROUTINE IS CALLED BY: rdcmps()   rdtree()   ufdbmp()   ufdump()`

This will cause doxygen to hyperlink that function name, so the user can click on it and go to the documentation for that function.

Emacs has C-x-l to lower-case text in a region, if you use emacs...

src/ufbdmp.f
```diff
@@ -1,6 +1,6 @@
 C> @file
 C> @author WOOLLEN @date 1994-01-06
-      
+C>
```
edwardhartnett, line 3: Ah ha! This will break things.

The reason is that there are actually two doxygen blocks. One, for the file, just has the author and date. The other, for the subroutine, has all the text.

If you put this 'C>' in, then doxygen will think that this is all file documentation, and there is no subroutine documentation.

So between the file documentation block and the first subroutine documentation block, there must be either a blank line, or a line with just 'C', but not a line with 'C>'.

This is subtle, and applies to many files.

src/ufbdmp.f
```diff
@@ -108,24 +109,23 @@
 C>    THIS SUBROUTINE STOPS THE SCROLL AND RETURNS TO THE CALLING
 C>    PROGRAM (PRESUMABLY TO READ IN THE NEXT SUBSET IN THE BUFR FILE).
 C>
-C>    THIS ROUTINE CALLS:        BORT     IBFMS    ISIZE    READLC
-C>                               RJUST    STATUS   UPFTBV
+C>    THIS ROUTINE CALLS:        BORT     IBFMS    ICBFMS   ISIZE
+C>                               READLC   RJUST    STATUS   UPFTBV
```
edwardhartnett, line 113: Similarly for lower-casing and putting parens.

src/ufdump.f
```diff
@@ -1,6 +1,6 @@
 C> @file
 C> @author WOOLLEN @date 2002-05-14
-      
+C>
```
edwardhartnett, line 3: Similar comment about the C> as above...

edwardhartnett (changes requested):

src/icbfms.f
jbathegit: I was going to get to those eventually, once I really go through each of the subprograms to fully "Doxygen"ate them ;-)  But if you want I'll go ahead and fix those now.

src/ufbdmp.f
jbathegit: This I had no clue about, so I'll go ahead and fix that as well, and while I'm at it I'll also go back and correct this in the other 5 source files that I've already fully documented.

jbathegit:
I will push another commit for the above requested changes.  Since I'm doing another commit for this anyway, is it OK if I go ahead and include the two small changes needed for #57, so we can get that one closed as well?

jbathegit:
OK done, so when you get a chance please (re)review! ;-)

jbathegit:
Just realized I'd missed icbfms.f in your initial request, so just pushed up another follow-on commit for that one as well ;-)

edwardhartnett (approved):

PR #59, opened 2020-11-09
Enable shared libraries and fix installation of non-standard prefix directories

markjolah:
Some small changes to the CMake which will improve portability of this library.  The changes should be transparent for downstream dependencies already using the exported CMake targets.

* Allow building either shared or static libraries.  
  * Static libraries are still built by default
  * Shared libraries can be built instead by setting the standard CMake `BUILD_SHARED_LIBS` option to 1.
    * See: [Cmake: BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)
* Conform installation to standard GNU prefix directory structure
  * Installation should respect the platform default installation directories as provided by [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)
  * Writing to directories outside of the standard locations is not portable.
  * Move the `include_${kind}` directories to `include/bufr_${kind}`.  This solves the problem of non-standard `include` directories in the prefix.
  * Use `CMAKE_INSTALL_LIBDIR` instead of `lib` as some platforms only have a `lib64` directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ include(GNUInstallDirs)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+option(BUILD_SHARED_LIBS "Enable building shared libraries" OFF)
+
 # Determine the kinds upfront (both src/ and test/ need them).
 # Both GNU and Intel support DYNAMIC_ALLOCATION
 list(APPEND kinds "4_DA" "8_DA" "d_DA")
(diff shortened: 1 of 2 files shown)


kgerheiser:
Looks good to me

aerorahul (approved):
Looks good.
Thanks @markjolah

kgerheiser (approved):

PR #60, opened 2020-11-10
Add Python API

### LINKED ISSUE #61: Add Python Api
Add Python Api from Jeff Whitaker. 
The python api is used in JEDI IODA projects.

aerorahul:
@jswhit2 developed a python API for NCEPLIBS-bufr some time ago.  It was transferred to JCSDA and it available [here](https://github.com/JCSDA-internal/py-ncepbufr).  The py-ncepbufr includes the NCEPLIBS-bufr source code.

This PR brings the Python API into the NCEPLIBS-bufr repository and builds it with cmake.

The Python API is used for BUFR data in JEDI application of IODA, and this will make it easier for testing, developing and using IODA converters for bufr data.

Fixes #61


edwardhartnett:
This is a great example of how we should use an issue to let us know what is coming and discuss. ;-)

@jbathegit did you know this was coming?

Some concerns:
* I see there are tests - which is good. How much of this code is tested?
* Are the python tests run in CI as well? They need to be.
* Jeff W. presumably wrote this on NOAA times, so we don't need extra legal statement about the code.
* Is there documentation?

edwardhartnett (changes requested):
See comments in main discussion...

aerorahul:
There are tests. They are not part of ctests as there is no baseline to compare. If just running a script is a test, I can add them in subsequent PR.

pdoc is used for creating documentation and is enabled with ENABLE_DOCS

I have no idea how much of the code is tested.

edwardhartnett:
Not sure what you mean by baseline. Do the tests need some data files to compare to?

How does pdoc fit into the existing doxygen build? We need integrated documentation, so we need pdoc to generate HTML is the appropriate place, and we need to reference that from the doxygen mainpage.

We need the tests to be turned on, and we need to know how much of this code is tested, and how to measure test coverage in python. (I've done it in the past but don't remember how...)

aerorahul:
There is test data. I can enable test running.

pdoc is what Jeff W. used to generate documentation for the Python API.  If you suggest we should replace that with Doxygen, fine.  Who will do that?

I am not sure how much of the code is tested.  I know how to measure test coverage, but I have no time to do so in this PR.

edwardhartnett:
Integrated documentation is a requirement. Please figure out how to integrate the documentation into doxygen. That means that a CMake build with ENABLE_DOCS=ON should generate HTML docs which include the pyhon docs. Let me know if you need help. You should not convert python documentation to doxygen.

Testing is a requirement, and there is a minimum level of testing (~70%). Therefore measuring test coverage is a requirement. If you need help, ask.

Preceding your PRs with an issue, especially about something as large as adding a new language API, will help you understand requirements before you submit work. But, basically, documentation and testing are now required for all new work on all NCEPLIBS projects.

If you don't have time to do the PR correctly, best take it down until you do.

aerorahul:
I am not sure when I will have time to integrate the documentation.  I could use some help, but I have no time to do that.
I am not sure what the current level of testing is (definitely not at 70%), and I don't think I will be able to implement or bring it to that level.  I could use help, but I have no time to do that.
Note: None of the NCEPLIBS projects have this in place, but I assume someone is working towards it.

I agree on preceding PR's with issues.

Issue #62 has been opened and can be used to track the progress of adding this new language API.

jbathegit:
@edwardhartnett To answer your question, no I didn't know this was coming.  This is all news to me as well.

That said, I think it will be great if we could add a Python API to the package, so I support the concept of doing this, and assuming we can integrate it cleanly as you described.

aerorahul:
@jbathegit 

I was hoping to bring this in and then incrementally clean it to fit in with the current layout, testing, coverage and documentation etc. 
That would have allowed us to use this without a loss of functionality.

However, to clean and prepare it to current layout, testing, coverage and documentation, is just not going to happen in any realistic timeline.  So, I think we will keep using the python API to bufr library that is outside this repo.

edwardhartnett:
No code will be added to any NCEPLIBS library without adequate testing and documentation.

It's a shame that more projects at NOAA are not following this common and time-tested rule of software engineering. If they were, not only would our software work better, but time would not be spent proposing and discussing the inclusion of work that does not meet these simple requirements.

Issue #17, opened 2020-10-08
measure test coverage


edwardhartnett:
What is the level of testing for this repo?

I see that there are tests and they are being run by github actions! Nice!

@jbathegit do you have a feel for how well this testing covers the code in this repo? Do we need more testing, and, if so, specifically what additional tests should we have?

jbathegit:
The 8 existing unit_test codes pretty-much cover all of the major subprograms and functions that would be used by 99% of users, so yes I think it's adequate, and I'll certainly let you know if/when any major functionality is added which would necessitate additional unit_test codes.

While any testing scheme could likely be improved, the subprograms which aren't directly called by the unit_test codes are lower-level routines which aren't intended to be directly called by users anyway.  Furthermore, most of those subprograms are at least implicitly tested anyway whenever the unit_tests are run, since the user-callable subprograms often call such lower-level subprograms within the library, and those subprograms in turn call other deeper-down subprograms, etc. to do a lot of the underlying work.

Bottom line, there are 300+ subprograms in the BUFRLIB, and I don't have the time or inclination to write a separate specific test case for every last one of them ;-)

edwardhartnett:
Awesome.

We need to measure test coverage and quantify this, and identify areas for additional testing.

edwardhartnett:
Attached is the lastest test coverage, just run this morning...
[coverage.tar.gz](https://github.com/NOAA-EMC/NCEPLIBS-bufr/files/5512332/coverage.tar.gz)

jbathegit:
Thanks very much for this @edwardhartnett , but is there any way you could generate this from an Intel DA build+test?  Two of the test codes (test_OUT_3.f and test_OUT_4.f) require a DA build of the library and are currently set up to only run on Intel.  So I'm guessing maybe you generated this latest version from a different build+test (maybe on Gnu?), because I don't see your latest graph showing any coverage for some of the subprograms that I just recently added new tests for in test_OUT_4.f (?)

Better yet, if you want to wait until after the new test_OUT_5.f is merged, that should show even better coverage.  And that one will run under both Intel and Gnu, and for both DA and non-DA (i.e. static) builds.

[...]

PR #65, opened 2020-11-10
include OUT_3 and OUT_4 test programs in GNU builds

jbathegit:
BINGO!!  This worked fine when I just tested it, so now it looks like we can have the full suite of DA test programs included for Gnu within our CI envrionment!

Fixes #11 

Part of #17


edwardhartnett (approved):

aerorahul:
if it works, then it works.

aerorahul (approved):

Issue #15, opened 2020-09-29
Add doxygen documentation


edwardhartnett:
As described here:
https://github.com/NOAA-EMC/NCEPLIBS/issues/121

edwardhartnett:
@jbathegit what is the sate of documentation for this code? I'm looking around and do see anything obvious...

jbathegit:
Sorry, I don't have any experience with doxygen, and I also don't have time to learn it right now.  That said, there are already docblocks inside all of the BUFRLIB subprograms, and there's also an online user guide at https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib.php

That's pretty-much all of the BUFRLIB documentation we have at this point, but feel free to take that and run with it if you like.

jbathegit:
Also, where did "Version 11.5.0" come from?  The next release version of BUFRLIB should be 11.4.0, unless I missed something(?)

edwardhartnett:
You have a release 11.4.0, so 11.5.0 seems next. ;-)

Who maintains the documentation here: https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib.php

Where is the source for this documentation?

Thanks!

jbathegit:
OK, clearly I've been a bit out of the loop, as 11.3.2 was the last release I was aware of.  But that's fine and no worries, and I appreciate all of the expertise and help from you and others to modernize the support and usefulness of this library for a wider audience.  My own particular area of knowledge centers around the legacy contents of the src and test subdirectories, mainly to support the WCOSS user community over the past many years.  But I realize we're collectively working towards a more modernized approach for all of the NCEP libraries using CMake, Doxygen and other tools, and I'm happy to support that wherever I can.

As for the web pages, I've been maintaining those pretty-much on my own for many years.  The latest source code is on the emcrzdm server, under the top-level directory /home/www/emc/htdocs/emc/pages/infrastructure/bufrlib/

edwardhartnett:
We are going to standardize on doxygen for all NCEPLIBS. We need to copy any useful info from the web pages to the doxygen docs. You can then continue to maintain the web pages if you wish, retire them, or add a link to the on-line doxygen docs. Whenever you add info to the web pages, make sure to add that to the doxygen docs as well.

[...]

PR #67, opened 2020-11-12
fixed doxygen docs for one subroutine

edwardhartnett:
Part of #15 

I needed a doxygen example for a presentation so I fixed a random subroutine. ;-)

### DIFF
--- a/src/getntbe.f
+++ b/src/getntbe.f
@@ -1,36 +1,24 @@
 C> @file
 C> @author ATOR @date 2007-01-19
 	
-C> THIS SUBROUTINE GETS THE FIRST LINE OF THE NEXT ENTRY IN
-C>   THE SPECIFIED ASCII MASTER TABLE B OR MASTER TABLE D FILE.  THIS
-C>   LINE CONTAINS, AMONG OTHER THINGS, THE FXY NUMBER CORRESPONDING TO
-C>   THIS ENTRY.
+C> This subroutine gets the first line of the next entry in
+C> the specified ascii master table b or master table d file. This
+C> line contains, among other things, the fxy number corresponding to
+C> this entry.
 C>
-C> PROGRAM HISTORY LOG:
-C> 2007-01-19  J. ATOR    -- ORIGINAL AUTHOR
+C> @param[in] LUNT integer: fortran logical unit number of ascii file
+C> containing master table b or master table d information
+C> @param[out] IFXYN integer: bit-wise representation of fxy number for
+C> next table entry
+C> @param[out] LINE character*(*): first line of next table entry
+C> @param[out] IRET integer: return code:
+C> -  0 = normal return
+C> -  -1 = end-of-file encountered while reading from LUNT
+C> -  -2 = I/O error encountered while reading from LUNT
 C>
-C> USAGE:    CALL GETNTBE ( LUNT, IFXYN, LINE, IRET )
-C>   INPUT ARGUMENT LIST:
-C>     LUNT     - INTEGER: FORTRAN LOGICAL UNIT NUMBER OF ASCII FILE
-C>                CONTAINING MASTER TABLE B OR MASTER TABLE D INFORMATION
-C>
-C>   OUTPUT ARGUMENT LIST:
-C>     IFXYN    - INTEGER: BIT-WISE REPRESENTATION OF FXY NUMBER FOR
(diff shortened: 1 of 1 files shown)


src/getntbe.f
```diff
@@ -1,36 +1,24 @@
...
-C>   THE SPECIFIED ASCII MASTER TABLE B OR MASTER TABLE D FILE.  THIS
-C>   LINE CONTAINS, AMONG OTHER THINGS, THE FXY NUMBER CORRESPONDING TO
-C>   THIS ENTRY.
+C> This subroutine gets the first line of the next entry in
+C> the specified ascii master table b or master table d file. This
+C> line contains, among other things, the fxy number corresponding to
+C> this entry.
 C>
-C> PROGRAM HISTORY LOG:
-C> 2007-01-19  J. ATOR    -- ORIGINAL AUTHOR
```
edwardhartnett, line 10: @jbathegit I am removing these PROGRAM HISTORY LOG sections when they contain nothing but the initial creation of the subprogram. That's already captured in the author and date tags, so no need to repeat it.

Whether we should continue updating these history logs is unclear. This is what github does already. Do we need to continue doing it? That question has been asked and I don't have any strong answer... Do you?

src/getntbe.f
```diff
@@ -1,36 +1,24 @@
...
-C>                NEXT TABLE ENTRY
-C>     LINE     - CHARACTER*(*): FIRST LINE OF NEXT TABLE ENTRY
-C>     IRET     - INTEGER: RETURN CODE:
-C>                       0 = normal return
-C>                      -1 = end-of-file encountered while reading
-C>                           from LUNT
-C>                      -2 = I/O error encountered while reading
-C>                           from LUNT
-C>
-C> REMARKS:
```
edwardhartnett, line 28: I am also removing all text elements like REMARKS:, ABSTRACT, etc. These may have been helpful in Fortran comment formatting, but are not needed for Doxygen documentation, which controls formatting without needing these.

src/getntbe.f
```diff
@@ -1,36 +1,24 @@
...
-C>    THIS ROUTINE CALLS:        BORT2    IGETNTBL IGETFXY  IFXY
-C>                               PARSTR
-C>    THIS ROUTINE IS CALLED BY: RDMTBB   RDMTBD   RDMTBF
-C>                               Normally not called by any application
-C>                               programs.
+C> This routine calls: bort2() igetntbl() igetfxy() ifxy parstr()
+C> This routine is called by: rdmtbb() rdmtbd() rdmtbf()
+C> Normally not called by any application programs.
```
edwardhartnett, line 21: Also note that whitespace is not significant for doxygen, so spacing things out to a certain column is not needed. We can use lists and tables to achieve a more organized effect when desire it.

jbathegit: Understood

jbathegit (approved):

edwardhartnett:
@jbathegit the date field should be the date the subroutine or function was first added, not the date it was most recently updated.

I will keep the PROGRAM HISTORY LOG from now on, even if it has only one entry. Should we lower-case it?

As noted in the call. the breif tag can be automatic everywhere. I will make sure that is turned on.

jbathegit:
IMHO, it's more useful and informative if the date is the date of the most recent update to the subprogram, not the date that the first version was written.  But you seem to be implying that the reverse is the industry standard, and if that's the case then I'll accept that approach and make the changes where needed - just let me know.

Yes, I'm fine with lowercasing the "PROGRAM HISTORY LOG" label going forward, and also the "THIS ROUTINE CALLS/IS CALLED BY" as well, though I may add <b> (i.e. bold) tags to them to allow them to still stand out a bit.

re: the brief tag, how does Doxygen know which part of the documentation to use for the brief?  Does it always just default to grabbing the first sentence or two following the `@file` section?

edwardhartnett:
Yes, the brief tag will be the first sentence of the documentation block, for files, subroutines, vars, etc. This works well and encourages good documentation practices.

Yes, the date field always refers to the date the function was added, that's how it's used everywhere. The most recent modified date is found in the github repo.

jbathegit:
OK, got it, though I will point out that when users are reading through the github-io documentation, they may not always also be looking back at the github repo ;-)

re: the brief tag, please let me know when you have the automatic capability turned on, and at that point I'll stop using it in all of the routines.

PR #72, opened 2020-11-13
Add macOS to Github Action

### LINKED ISSUE #70: CI needs to test on macOS as well as ubuntu
CI needs to test on macOS as well as ubuntu. This is related to #68

kgerheiser:
Fixes #70

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,10 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-10.15,  ubuntu-20.04]
     env:
       FC: gfortran-9
       CC: gcc-9


.github/workflows/main.yml
```diff
@@ -3,7 +3,10 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-10.15,  ubuntu-20.04]
```
edwardhartnett, line 9: So easy! What other OS flavors are available?

edwardhartnett (approved):

.github/workflows/main.yml
kgerheiser: Ubuntu 16, 18, 20

macOS 10.15, 11.0

Windows Server

edwardhartnett: Seems like we should use all of them, except Windows.

aerorahul:
can we test with `clang` too please on a macOS?

aerorahul:
The DA applications, believe it or not, tests with `clang`

aerorahul:
This is not right.
there is no sudo apt-get on a macOS.

aerorahul:
Ok. Well you don't have that issue.  I will need to create a separate workflow run for macOS, because we are going to need sudo apt-get and brew for the python PR.

aerorahul:
I was able to run the CI for the python PR.  All tests are passing for both macos and ubuntu. phew!

PR #73, opened 2020-11-13
doxygen fix

edwardhartnett:
Part of #15 

Convert one function to explore use of brief tag. It's needed at the file level, but not at the subroutine level.

### DIFF
--- a/src/adn30.f
+++ b/src/adn30.f
@@ -1,54 +1,40 @@
 C> @file
+C> @brief Converts a descriptor from its bit-wise (integer) representation to
+C> its five or six character ascii representation.
 C> @author WOOLLEN @date 1994-01-06
+
+C> @author WOOLLEN @date 1994-01-06
+C>
+C> This function converts a descriptor from its bit-wise
+c> (integer) representation to its five or six character ascii
+C> representation.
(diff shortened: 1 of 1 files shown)


edwardhartnett:
With this, I see the description in the file list:

image

edwardhartnett:
One problem (?) is that a separate brief tag is needed for file and subprogram. Normally this makes sense because one file holds multiple subprograms. So the file level one summarizes what's in the file, and each subprogram summarizes what's in the subprogram.

But in this case there is one subprogram per file...

jbathegit:
Yeah, I thought about that too, but what I'm going to do for this is to have a slightly more general description at the file level vs. at the subprogram level, the idea being that when everything's displayed in "Files" we want the user to be able to quickly determine where they need to drill down to find what they're looking for.  Hopefully you'll see what I mean when I submit my next PR ;-)

jbathegit (approved):

jbathegit:
Thanks for playing around with this to help figure this out!

edwardhartnett:
I do agree that the list of files looks much nicer and more useful when each file has a brief description next to it.

PR #82, opened 2020-11-20
include contents of utils directory in Doxygen output

### LINKED ISSUE #26: README.md in utils directory should be added to doxygen docs
README.md in utils directory should be added to doxygen docs

jbathegit:
Fixes #26 

Please see https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/index.html for an example of what this will eventually look like on Github.io  A new Related Page "Sample Utilities" is now generated, and I've also added `@file` and `@brief` tags to all of the associated *.f files so that the description for each one shows up under the Files tab.  Furthermore, in the README.md I've also added links from each utility to that source documentation under the Files tab, which in turn links to the actual source file just like for all of the BUFRLIB subprograms.

FYI, I also had to change around some punctuation in the README.md file, because Doxygen and GitHub have slightly different rules about what types of markdown extensions they will and won't accept. So, for example:
- I had to change the ``` delimiters to ~~~ surrounding all of the code blocks.  The former only works in GitHub, but the latter works in both.
- I also had to add `<div>` HTML tags to each section, because Doxygen doesn't recognize the ### as a target for a relative link from the summary table at the top of the file.


jbathegit:
So as you can see I've gone ahead and added another commit to temporarily turn off the ENABLE_DOCS check in the GitHub builds, so we can continue to commit to `develop` while @aerorahul considers options to fix the associated Python issues.  Then at that point we can go back and re-enable the check in main.yml

Still waiting for the macos-10.15 CI tests to run, then will let you know so you can review this PR...

aerorahul:
I will explore `doxypy`.  `pdoc` and `doxypy` both generate html documentation, albeit the looks are different. 
I agree in the long run since we have chosen the doxygen look, we should use `doxypy`. I chose to keep with `pdoc` because that was what was available.

I'll chip away at it when I have time or someone from EIB should look to do it if they want to put a timeline.

jbathegit:
Thanks @aerorahul, that sounds good!

In the meantime, do you or @edwardhartnett have any thoughts on why it's taking so long for the two macos-10.15 tests to start?  They've been queued up and waiting to run for over 20 minutes!?  The ubuntu tests both successfully finished long ago.

jbathegit:
OK, so it finally ran, but errored out with a different python error, something about "building wheel for markdown"(?)  This is only happening on mac-os.  Everything's fine on ubuntu.

See https://github.com/NOAA-EMC/NCEPLIBS-bufr/pull/82/checks?check_run_id=1432883289 for more details, specifically lines 57 and 140 of the install-deps output

@aerohul - any thoughts?  I'd really like to get this working again so we can resume commits to the develop branch.

aerorahul:
We need to disable installing python pdoc in the install_deps.
Just turning `-DENABLE_DOCS=OFF` is not really the issue if the CI is trying to install markdown in the install_deps step.

jbathegit:
OK, so is what you're suggesting to remove the `pip3 install pdoc` step from main.yml?  And should I also remove the same step from the ubuntu block for the time being, even though it seems to work fine there?

Please confirm and I'll push up another commit - thanks!

edwardhartnett:
You only need to remove it on macos...

jbathegit:
OK just pushed up another commit (fingers crossed ;-)

aerorahul:
I have no idea why the macOS runners are "waiting to start"

jbathegit:
Me neither, but this is getting a bit ridiculous ;-)

jbathegit:
Finally - DONE!

utils/README.md
```diff
@@ -1,15 +1,21 @@
-# BUFR Utilities
 
-This is a collection of commonly used BUFR utilities:
-[readbp](#readbp) - Read prepbufr file and print reports
-[readmp](#readmp) - Read any BUFR file and dump contents of each subset one at a time
-[binv](#binv) - Print BUFR file inventory by message type
-[sinv](#sinv) - Print inventory of satellite data by platform and instrument type
-[cmpbqm](#cmpbqm) - Print inventory of prepbufr observations by variable, report type and quality marks
+#### Sample Utilities
+@brief Collection of commonly-used utilities based on the library
```
aerorahul, line 3: the `@brief` shows up on page.

utils/README.md
```diff
@@ -1,15 +1,21 @@
...
+| [sinv](#sinv) | Print inventory of BUFR satellite data file by platform and instrument type |
+| [cmpbqm](#cmpbqm) | Print inventory of observations from prepbufr file by variable, report type and quality mark |
+
+<br>
 
 ---
 
-## readbp
+<div id="readbp"/>
+### readbp
```
aerorahul, line 18: Its unable to interpret `###`

aerorahul: here and in every heading.

aerorahul (commented):
doesn't seem right. 
Take a look at the page as rendered on github.
https://github.com/NOAA-EMC/NCEPLIBS-bufr/tree/jba_READMEutils/utils

utils/README.md
jbathegit: Yeah, I know but we need it for Doxygen, and I figured it didn't detract too much from the overall readability.  As I noted above, Doxygen and GitHub have different rules about what markdown extensions they will and won't accept, so in some places I had to compromise a little to make it work in both places.

aerorahul: Yeah. I don't like that.  One has to click to go to doxygen page or build it.  This is the front page

utils/README.md
jbathegit: Hmm, if I just type `### readbp` at the beginning of a new line right here in this GitHub window, it works fine:

### readbp

but for some reason it doesn't work when rendered on github.com.  Maybe it needs an extra blank line between the `<div>` tag and itself?

jbathegit: Yep, that's exactly what it was (grrr!!).   OK, I'll push up yet another commit to fix this, and maybe I'll go take a nap while I'm waiting for mac-os to run again ;-)

aerorahul:
much better.
Can we target that `@brief`.
It just looks sloppy.

jbathegit:
I'll see what I can do, but again I was trying to make things work for two different and (in their own ways) very picky interpreters.

aerorahul:
Why do we need to render this text into a doxygen html page?
If we do, then why do we care about also having a github README.md document?
Just thinking out loud.

jbathegit:
As far as I'm concerned, the Doxygen docs are more important than the utils/README.md file.  But that said, the way it is now allows the content to be rendered in both places, even if not perfectly.  And in my opinion that's the most important thing - to get the documentation in front of as many eyeballs as possible.  But if I had to choose one to support over the other, my vote would be the Doxygen, because that's going to have a lot of additional NCEPLIBS-bufr documentation that's never going to appear in any README.md file.

Anyway, back to your question about the `@brief` tag, I've tried every idea I could find online, and I still don't have a solution.  I can hide it from GitHub by encapsulating in `<!-- -->` (i.e. HTML comments) or even `<!--- --->` which I also saw mentioned in several places, but unfortunately that also hides it from Doxygen.

Is this really that big of a deal?  In other words, would you mind going ahead and approving these commits for now, and if you decide that the `@brief` tag is still bothering you, maybe just open this up as a separate new issue that we can try to revisit again later?

jbathegit:
## Sample Utilities
@brief Collection of commonly-used utilities based on the library

jbathegit:
per the previous comment, I could make the "Sample Utilities" text larger if that helps(?)

edwardhartnett:
I agree that the doxygen docs are the canonical docs, and that looking at the readme via github is not the way people should be reading the docs.

Perhaps it would be best to abandon the idea of using GitHub to view the docs, and perhaps even move the file to the docs directory to remove the idea that the readme exists to document the directory it is in.

On NCEPLIBS we'll support the one README file in the main directory, but that should be it. We don't need to multiply the forms of documentation that we maintain. The doxygen is enough. When we can get it to generate the readthedocs documentation, then doxygen will be the only documentation we maintain.

So I would say that removing the brief to improve the appearance of any doc file in GitHub is not a requirement, unless we're talking about the README.md in the root directory. Jeff has found a great way to integrate .md files into doxygen, but that does not mean they also have to render perfectly in GitHub.

aerorahul:
@jbathegit @edwardhartnett 
I suggest we reduce the `READMEutils.md` down to a list of tools and a one-liner of what each of the utility does.
The documentation can be moved inside the fortran code or in its own documentation to eliminate duplication.
This so-called documentation/usage came from @jack-woollen.
The `READMEutils.md` does not have to be part of doxygen documentation.  It can provide a link to the doxygen documentation on RTD of the utilities page.

edwardhartnett:
@jbathegit I believe you need to merge this before the CI on my PR will pass.

@aerorahul we will let Jeff organize the documentation. Other than the main README of the project, the readmes can be whatever Jeff needs to get the doxygen output the way he wants it.

jbathegit:
> @jbathegit I believe you need to merge this before the CI on my PR will pass.

@edwardhartnett, actually, I'm still waiting on either you or @aerorahul to review and approve this PR before I can merge it! ;-)

aerorahul:
@jbathegit 
Can we remove the contents of the README.md in utils and reduce it to a table.
The usage can be included in the utilities themselves.  
This will ensure there is no duplication and the github readme is a mere listing of the utilities.

aerorahul (approved):

PR #85, opened 2020-11-25
Allow for user python installation to be used for the Python API build and install process on Mac OS

### LINKED ISSUE #84: Python API does not allow the use of a user python installation on Mac OS
If a user has installed miniconda as an example, the cmake FindPython script, by default, prefers the macos framework python installation over any other, even when the PATH is set to point to the user's miniconda installation.

This breaks if for instance the frameworks python is version 3.7 and the users miniconda version is 3.8. (This is how the issue was found.) The setup.py build procedure for the Python API builds a shared library file with the python version embedded in its name (eg, `_bufrlib.cpython-37-darwin.so` when python3.7 is used to run setup.py). When the frameworks version is 3.7 (used for the build) and the miniconda version is 3.8 (attempting to import ncepbufr), the shared library file cannot be loaded because miniconda expects the file name to be `_bufrlib.cpython-38-darwin.so` (with a 38 instead of a 37 embedded in it's name).

It would be nice if the build, install process for cmake would use the user's python installation instead of the macos framework so that share library file name is consistent with what the user wants to use.

srherbener:
This PR introduces a change to the cmake configuration that allows the python API build and install process to use a user's python installation instead of forcing the macos framework python installation to be used. This is accomplished through a control variable that the cmake FindPython script recognizes which forces the framework installation to be searched last instead of first for available python installations on the platform.

This fixes #84


srherbener:
In case you are wondering who I am, my name is Steve Herbener and I'm a UCAR employee working on the JEDI projects in JCSDA. I work regularly with EMC on collaborative JEDI code development.

aerorahul (approved):
Thank you for reporting the bug and providing a solution.

jbathegit (approved):

PR #88, opened 2020-12-18
add utility to dump table

### LINKED ISSUE #87: add utility to dump table.
Add another utility from @jack-woollen that dumps the bufr table from a file.

aerorahul:
add utility `gettab.x` to print bufr table from the file.
closes #87


edwardhartnett:
Please do not commit this change without a test.

If  there is not a test, how do I have any idea this code works? How do I know you won't break it tomorrow?

aerorahul:
How do you test something that is printing to stdout?

Are there tests for other utilities in this repo?

edwardhartnett:
I've described a way of testing over in the PR. It's how we test ncdump in the netcdf-c package, for example.

@jbathegit can fill you in on whether any other utilities are tested yet. If not, then this can be the first.

aerorahul:
A test is added to the `gettab.x` utility.  Test output is also added to check against. 
The utility calls the function `dxdump`.

utils/gettab.f90
```diff
@@ -0,0 +1,28 @@
...
+endif
+
+open(lunit,file=trim(adjustl(finput)), status='old', form='unformatted')
+call openbf(lunit, 'IN', lunit)
+call dxdump(lunit, 6)
+call closbf(lunit)
+close(lunit)
+
+stop
+end
```
rmclaren, line 28: Fix indentation. Indent everything inside of `program` section.

rmclaren: Last line should be `end program gettab`, also file should end with t a blank line.

utils/gettab.f90
```diff
@@ -0,0 +1,28 @@
+!> @file
+!> @brief Read BUFR file containing embedded DX tables,
+!> and print the table to stdout
+
+program gettab
```
rmclaren, line 5: consider renaming to get_table, or gettable or something like that.

utils/readmp.f90
```diff
@@ -0,0 +1,46 @@
+!> @file
+!> @brief Read BUFR file containing embedded DX tables, and print
+!> each report one at a time
+
+program readmp
```
rmclaren, line 5: Consider renaming to something that makes sense... read_**something**

test/test_gettab.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
edwardhartnett, line 1: Comments please. What is this file, who wrote it, and what's it for?

utils/gettab.f90
```diff
@@ -0,0 +1,28 @@
+!> @file
+!> @brief Read BUFR file containing embedded DX tables,
+!> and print the table to stdout
```
edwardhartnett, line 3: Author and optionally date please.

utils/readmp.f90
```diff
@@ -0,0 +1,46 @@
...
+      close(lunit)
+      stop
+    endif
+  enddo
+enddo
+
+call closbf(lunit)
+close(lunit)
+
+stop
```
rmclaren, line 45: Same comments as the last one.

utils/readmp.f90
```diff
@@ -0,0 +1,46 @@
+!> @file
+!> @brief Read BUFR file containing embedded DX tables, and print
+!> each report one at a time
```
edwardhartnett, line 3: Author and date tags?

edwardhartnett (changes requested):

utils/readmp.f90
rmclaren: The version control system keeps track of this information... there really shouldn't be a need to have to annotate each file this way.

edwardhartnett: Please add author tags to all code in all NCEPLIBS projects. Date tags are optional but encouraged.

The repo is not sufficient for this purpose, as evidenced by the fact that I have over 200K lines of Fortran/C code, some 30+ years old, all added to GitHub recently, with incorrect attribution. Had those programmers not properly documented their work, we would now have no idea who wrote what. 

When a scientist publishes a paper, their name goes at the top, which documents their involvement, gives them proper credit, and encourages them to take extra care that the paper is correct, as it reflects directly on them. 

For the same reasons, all NCEPLIBS codes will contain proper attribution.

test/test_gettab.sh
aerorahul: done.

utils/gettab.f90
aerorahul: done.

utils/readmp.f90
aerorahul: done.

utils/gettab.f90
aerorahul: I would like to rename this to something that makes sense.  Also all the cryptic functions and subroutines.
 However, I am not the one who wrote this utility and it is under circulation for decades without being version controlled or packaged with the library.
I am trying to bring these orphans into one roof (or repo!) to help users (and developers) alike.

utils/readmp.f90
aerorahul: done per the decree.

utils/gettab.f90
aerorahul: done.

utils/readmp.f90
aerorahul: It should be `read_message`, ideally.

utils/gettab.f90
aerorahul: this should be `get_table`, ideally.

aerorahul:
On further thought, these utilities should have a prefix `bufr_` or something like that to distinguish them as being `bufr` utilities.

`get_table` and `read_message` are too generic, and when installed on system paths e.g. `/usr/local/bin`, could cause headaches.  Not to mention, they don't have a `usage` or anything like that to instruct how to use them or what they are.

Something to think about for the `bufr` developers.

utils/readmp.f90
jack-woollen: how about bfdump? (witness ncdump)

utils/readmp.f90
jack-woollen: Appreciate the comments and examples, thanks. I should be taking care of this.

utils/gettab.f90
jack-woollen: maybe 'bfdump -table'.  see comment for readmp.

edwardhartnett:
So sounds like this should not be merged until we sort out the names, and also add usage messages...

utils/readmp.f90
edwardhartnett: Also, as this is fortran code, it should contain doxygen documentation.

edwardhartnett (changes requested):
Issues outstanding:
* names of utilities
* doxygen documentation

aerorahul:
As this task has now evolved from adding an orphan utility to providing proper names of the utilities and comprehensive documentation and testing, it is beyond the scope of this PR.

Before this PR:
- there were no tests to the utilities
- there was no documentation of the utilities (Doxygen does not look in this directory).

I have added documentation to the utility added in this PR `gettab.x`.
I have reverted `readmp.f90` to the original `readmp.f`.
I do not wish to take up modernizing pre-historic code (even in incremental steps) in this library.

If it would please the developers, accpet this PR based on what it is; add a utility.
The utility (from @jack-woollen) provides a test and documentation.

aerorahul:
> So sounds like this should not be merged until we sort out the names, and also add usage messages...

Ok. Can this be addressed in a separate issue/PR.

rmclaren (approved):

utils/readmp.f90
aerorahul: done.

aerorahul:
@edwardhartnett need your approval before proceeding.

edwardhartnett:
I was hoping that @jbathegit would take a look at this...

aerorahul:
> I was hoping that @jbathegit would take a look at this...
Ok.  What in particular is the sticking point right now for you?

edwardhartnett (approved):

jbathegit:
@edwardhartnett and @aerorahul, I've been on leave the past several weeks, which is why I haven't chimed in until now.  That said, I have no issues with the content here, and I'm also fine with changing the name to something else as part of a separate PR, though I'm not a particular fan of `bfdump` since that's very similar to the library subroutine name `ufdump` and might lead to some confusion.  I'm actually fine with just leaving it as `gettab`, but I also wouldn't be opposed to adding a `bufr_` prefix to it or any of the other existing utility names.

It's true that none of the other utilities have any tests associated with them yet, and that would be a nice addition to the package.  However, I didn't author any of those utilities myself, so I'm not volunteering to take on that task.

jbathegit (approved):

edwardhartnett:
@jbathegit all utilities in NCEPLIBS-bufr need tests. For those that don't have them, they must be written. Since you are the code manager, you have already been volunteered for this task. ;-)

As we discussed before, the reason for you to not let untested code be merged is that you are responsible for writing tests for all untested code. So if someone merges untested code, they are directly adding to your workload. Bear this in mind for all PRs.

@aerorahul did include a test for his change, so now we just need tests for everything else. ;-) As @aerorahul points out, the utilities also need to be checked for code coverage, and I add them next time I run the code coverage numbers.

All released code must be tested - every Fortran subprogram, every utility, every shell script. **_Everything_**. It's a firm requirement from this point forward. Where those tests are currently missing, the code manager is responsible for adding them. Please add issues to add a test for each specific utility.

I hope we did not do anything incorrectly in your absence. Take a look at what was merged (including the test) and let us know if you have any issues with it...

jack-woollen:
@jbathegit  I write whatever tests are needed for the utils i made.

jack-woollen:
@jbathegit That is to say, I will write any test needed. Haven't done it yet.

jbathegit:
@edwardhartnett - I hear what you're saying, but those utilities were added to the package before the test code mandate, and not by me.  At least in the past, I wasn't the only person who was approving and merging updates to the main branch, even though I may have been the only person officially listed as a code manager.  So I think it's a bit unfair to now say that I'm responsible for writing tests for all such legacy utilities in the repository, and thus I'm very grateful to @jack-woollen for his offer to help out with this!

Going forward, what you're saying makes sense from a theoretical standpoint, but in practical terms this is a huge responsibility for one person.  I understand and agree with the "no test, no merge" rule and am happy to help enforce it going forward, but it can be hard for me to judge what does and doesn't constitute adequate testing since I don't have any way to run gcovr myself. and thus I can't easily check code coverage without help from you or someone else.  Furthermore, as I've noted before, this is a library with an enormous amount of code and lots of visibility throughout NCEP and beyond, so I've always relied on help from others (mostly Jack, but also Dennis K. and others at times in the past) to be able to manage and support it.  I know you've said that in NCEPLIBS we work together as a team, but in reality most responsibility seems to always fall on the code manager, and again that's a huge responsibility for this particular library, especially now that we have additional Python interfaces, CMake scripts, and other technical content in which I have little to no personal expertise of my own.  So is there any chance we could have an additional person to officially serve as "co-" code manager for NCEPLIBS-bufr going forward?

Again, many thanks to @jack-woollen for stepping up to help with writing tests for the other existing utilities!

edwardhartnett:
@jbathegit everything we release must be tested. Where tests don't exist, we must write them. The fact that we did not originate the code does not mean the code does not need to be tested. ;-)

Yes, a lot of this comes down to the code manager, which is a great incentive to only accept PRs that include full testing.

I agree that you must be able to run gcovr yourself. One way to do that would be to get you a NOAA linux or mac laptop, on which you could build NCEPLIBS-bufr with GNU tools, as I do. Would that work for you?

Another way to ensure this is to carefully inspect all test code and just see what is being tested. Is it barely testing anything? Are error conditions and bad input tested? Are all major features part of the test? Are the tests documented?

Please work with @jack-woollen to get all his scripts tested. If you can identify other programmers who have contributed untested code, we can ping them and get them to write tests. Every programmer should be happy to take responsibility for writing tests for their own code. If we need to get management buy-in for particular programmers, let me know.

This problem is general to NCEPLIBS, and NCEPLIBS-bufr is not the only one that needs help. ;-) This will not all happen in a week or a month, it will take us much of 2021 to get these codes tested. I hope that we can get you some help with this, but it will not be immediate.

PR #90, opened 2020-12-22
Split bufr file into subsets

### LINKED ISSUE #89: add utility to split bufr files by subsets.
The bufr files in the "dump" are a result of the `dumpjb` step.  Several bufr operations occur in this step; one of which is combining bufr files from the "tank" into a single bufr file in the "dump".  This is done by the obsprocessing code `combfr.f`.

`combfr.f` is a generic bufr utility that concatenates a number of bufr files into a single one and would ideally reside in NCEPLIBS-bufr. 

This issue relates to adding a utility to split the "dump" bufr file into its subsets, so that they can be read in by the DA framework of JEDI.
When the DA framework reads bufr files straight from the "tank", this utility won't be used (although may still be useful).

aerorahul:
This PR:
- addes a utility to split a bufr file (e.g. satwndbufr) into individual subsets
- closes #89 
- is built on top of PR #88 See [this](https://github.com/NOAA-EMC/NCEPLIBS-bufr/compare/feature/more_utils...feature/split_by_subset) for comparison with PR #88

This utility is the opposite of the `combfr` utility used in the `dumpjb` job that combines/concatenates the bufr files int a single file.

A test is added.
Test output files will need to be uploaded to the ftp site.

To run the test locally (until the test files are placed on the ftp site), run the `split_by_subset.x` on `testfiles/data/satwndbufr` and copy the resulting files in `testfiles/tesoutput/satwndbufr_split` and copy the resulting `NC??????` files into `testfiles/testoutput/satwndbufr_split` directory.

Then re-run the test.


rmclaren (approved):

jack-woollen (approved):
Reviewed and approved.

aerorahul:
@jbathegit 
Could you please add the new test files required for this PR in `bufr.tar` and upload it to the ftp site?
I have created an updated `bufr.tar` and it is on Hera at: `/scratch1/NCEPDEV/stmp2/Rahul.Mahajan/test/ObsProc/test_data/bufr.tar`

Once the files are in place on the ftp site, I can re-trigger the CI.  The test associated with this PR will fail until those files are part of the `bufr.tar`

jbathegit:
Sorry, I don't have access to hera.  Any chance you could put the new `bufr.tar` file somewhere on the wcoss (venus or mars)?  Or maybe on a public ftp server somewhere, or even just somewhere on emcrzdm and then I can move it over to the correct subdirectory for curl to pick it up?

I presume that the files you want to add to `bufr.tar` are the files in `testfiles/testoutput/satwndbufr_split`, correct?  If so, that's fine, but then for consistency's sake (and to avoid confusion in the future), I think it would be nice to also have `1bamua.table.out` (from the previous gettab.f90 test) to also be moved into the `testfiles/testoutput` directory within the `bufr.tar` file, rather than continuing to keep it as a separate standalone file within the `test/testoutput` subdirectory of the distribution package.  That way we avoid having two separate `testoutput` directories in two separate places, again with the idea of avoiding confusion down the road.

aerorahul:
> Sorry, I don't have access to hera. Any chance you could put the new `bufr.tar` file somewhere on the wcoss (venus or mars)? Or maybe on a public ftp server somewhere, or even just somewhere on emcrzdm and then I can move it over to the correct subdirectory for curl to pick it up?
> 
> I presume that the files you want to add to `bufr.tar` are the files in `testfiles/testoutput/satwndbufr_split`, correct? If so, that's fine, but then for consistency's sake (and to avoid confusion in the future), I think it would be nice to also have `1bamua.table.out` (from the previous gettab.f90 test) to also be moved into the `testfiles/testoutput` directory within the `bufr.tar` file, rather than continuing to keep it as a separate standalone file within the `test/testoutput` subdirectory of the distribution package. That way we avoid having two separate `testoutput` directories in two separate places, again with the idea of avoiding confusion down the road.

@jbathegit My WCOSS account is locked and my emcrzdm password has expired.  Once I have that restored, I can put it on WCOSS or emcrzdm.

I can move the `1bamua.table.out` file from the previous test in the `testfiles/testoutput`.

aerorahul:
@jbathegit 
I have uploaded the new `bufr.tar` on emcrzdm at `/home/www/emc/htdocs/gc_wmb/rmahajan/bufr.tar`
Could you please replace the one on the ftp site with this one?
Thanks!

jbathegit:
@aerorahul

OK, it's done.  You can go ahead and restart the CI test.

aerorahul:
> @aerorahul
> 
> OK, it's done. You can go ahead and restart the CI test.

Thanks @jbathegit 
All tests are now passing. Good to go.

jbathegit (approved):

test/test_scripts/test_split_by_subset.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#> @file
+#> @brief Test script to test split_by_subset.x utility
```
edwardhartnett, line 3: Tests do not have to have doxygen, and should not. (For future reference.)

utils/split_by_subset.f90
```diff
@@ -0,0 +1,104 @@
+!> @file
+!> @brief Read BUFR file messages, collating them into output files by message type/subtype (eg NC001002, aka subset type)
+!> This is like the opposite of combfr.f which concatenates BUFR messages from listed input files
+!> @author WOOLLEN
```
edwardhartnett, line 4: Date tag is also helpful when new things are added.

Issue #81, opened 2020-11-18
Possible Standards Violation in moda_mstabs.F


mathomp4:
All,

Recently I tried to compile the NCEP_bufr library in GEOS which led to an [issue with compiling with GCC 10](https://github.com/GEOS-ESM/NCEP_Shared/issues/16) with `moda_mstabs.F`:
```
[  0%] Building Fortran object src/Shared/@NCEP_Shared/NCEP_bufr_r4i4/CMakeFiles/NCEP_bufr_r4i4.dir/moda_mstabs.F.o
/discover/swdev/mathomp4/Models/GEOSgcm-with-bufr-GCC10/GEOSgcm/src/Shared/@NCEP_Shared/NCEP_bufr/moda_mstabs.F:39:44:

   39 |    CHARACTER*4, BIND(C) :: CBBW(MXMTBB)
      |                                            1
Error: BIND(C) Variable ‘cbbw’ at (1) must have length one
/discover/swdev/mathomp4/Models/GEOSgcm-with-bufr-GCC10/GEOSgcm/src/Shared/@NCEP_Shared/NCEP_bufr/moda_mstabs.F:42:48:

   42 |    CHARACTER*120, BIND(C) :: CBELEM(MXMTBB)
      |                                                1
Error: BIND(C) Variable ‘cbelem’ at (1) must have length one
/discover/swdev/mathomp4/Models/GEOSgcm-with-bufr-GCC10/GEOSgcm/src/Shared/@NCEP_Shared/NCEP_bufr/moda_mstabs.F:41:46:

   41 |    CHARACTER*8, BIND(C) :: CBMNEM(MXMTBB)
      |                                              1
Error: BIND(C) Variable ‘cbmnem’ at (1) must have length one
/discover/swdev/mathomp4/Models/GEOSgcm-with-bufr-GCC10/GEOSgcm/src/Shared/@NCEP_Shared/NCEP_bufr/moda_mstabs.F:37:45:

   37 |    CHARACTER*4, BIND(C) :: CBSCL(MXMTBB)
      |                                             1
Error: BIND(C) Variable ‘cbscl’ at (1) must have length one
/discover/swdev/mathomp4/Models/GEOSgcm-with-bufr-GCC10/GEOSgcm/src/Shared/@NCEP_Shared/NCEP_bufr/moda_mstabs.F:38:47:

   38 |    CHARACTER*12, BIND(C) :: CBSREF(MXMTBB)
[...]

PR #94, opened 2021-01-11
Use GCC-9 and GCC-10 in Github Action

### LINKED ISSUE #93: Test with gfortran 10
We need to test with gfortran 10. Should be possible on Github actions.

kgerheiser:
Add GCC-10 to build matrix

Fixes #93

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,10 +7,8 @@ jobs:
     strategy:
       matrix:
         os: [macos-10.15,  ubuntu-20.04]
-    env:
-      FC: gfortran-9
-      CC: gcc-9
-
+        compiler: [gcc-9,  gcc-10]
+        
     steps:
 
     - name: install-deps
@@ -29,9 +27,13 @@ jobs:
           pip3 install setuptools
           pip3 install numpy
           pip3 install netCDF4
-          sudo ln -sf /usr/local/bin/gfortran-9 /usr/local/bin/gfortran
+          if [[ ${{ matrix.compiler }} == "gcc-9" ]]; then
+            sudo ln -sf /usr/local/bin/gfortran-9 /usr/local/bin/gfortran
+          elif  [[ ${{ matrix.compiler }} == "gcc-10" ]]; then
+            sudo ln -sf /usr/local/bin/gfortran-10 /usr/local/bin/gfortran
+          fi
         fi
-
+        
     - name: checkout
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
Looks good.

jbathegit:
Thanks very much for this @kgerheiser , but I'm a little surprised by one thing here.  Specifically, it looks like you had a successful gcc-10 build even though you're testing on the develop branch, whereas @mathomp4 wasn't able to do that, and which is why he opened #81 in the first place.  In other words, you're not testing on the jba_mstabs branch that I created to fix that issue that he reported.  So now I'm curious how you were able to do that, or what you may be doing differently from what he was doing?

@mathmop4, were you doing your own compilation outside of our CMake scripts, perhaps with some different compiler flags than we use in our standard builds?  Again, I'm wondering why/how @kgerheiser apparently had no problems with the GNU v10+ compiler even though he wasn't testing against my new branch(?)

kgerheiser:
Yes, building with GCC 10 just worked for me. I didn't really look at the original error.

mathomp4:
> Yes, building with GCC 10 just worked for me. I didn't really look at the original error.

It's possible @kgerheiser builds bufr with DYNAMIC_ALLOCATION? Or perhaps that is the default in whatever version he has? If so, you avoid the bug.

jbathegit:
OK, then it sounds like maybe we need to add testing for STATIC as well as DYNAMIC builds, in order to try to reproduce this error.  I hadn't realized this before, but it looks like the static builds are only being tested for the Intel compiler!?  Here's a snippet from `test/CMakeLists.txt`:

<pre>
# Only Intel supports STATIC_ALLOCATION
if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
  list(APPEND test_kinds "4")
endif()
</pre>

Does anyone know why this filter is here?  Unless I'm missing something, we should be testing all builds (both STATIC and DYNAMIC) for all compilers, not just Intel.  The only exceptions are `test_OUT_3.f` and `test_OUT_4.f`, which can only be run on DYNAMIC builds because they specifically test features of dynamic allocation.  But I can't think of any reason why we shouldn't be testing all of the other STATIC builds on all compilers.

**Please understand that when we're talking about STATIC vs. DYNAMIC here, we're talking about how arrays are allocated internally within the library, not about how the library is linked to an application program!**  So again, at least in my mind there's no reason we shouldn't be testing the STATIC builds as well as the DYNAMIC builds on all platforms (again, for all test programs except for `test_OUT_3.f` and `test_OUT_4.f` ;-)

jbathegit:
Also, @kgerheiser, I noticed that there are two checks from yesterday which are still hung in a _"Waiting for status to be reported"_ state.  Do you have any idea what's going on here?

kgerheiser:
Yes, since the checks changed, they need to be updated in the repository settings

kgerheiser:
I've updated the branch settings. It can be merged now. I think we should merge this and then deal with the other issues separately.

jbathegit:
OK, I'll go ahead and merge your main.yml change, and then we can deal with the static vs. dynamic as a separate issue.

Issue #77, opened 2020-11-17
Transition to DA build only


edwardhartnett:
Currently twe build a static and a dynamic allocation memory model.

Let's eliminate the static build and make the dynamic build the only build.

edwardhartnett:
@mathomp4 you mention in #81 that you use the static build. Is there a reason why you prefer the static instead of the dynamic allocation build?

mathomp4:
> @mathomp4 you mention in #81 that you use the static build. Is there a reason why you prefer the static instead of the dynamic allocation build?

I prefer it only because that is how it was built in our CVS GNUmake build that we are moving to Git. And my mandate was to make sure all the flags were *exactly* the same (so we could get the same answer).

My guess is the new BUFRLIB was brought in and "things" were done until it worked (this was after I moved from maintaining the CVS/GNU build to our current Git/CMake). Maybe they didn't know about the dynamic allocation build? I'll propose we move to it, but I don't know how to evaluate if doing so changes results.

Note we only saw this because (until now) our main GCM didn't build this library as it didn't need it. But recent work has discovered it will be needed, so it was added and our CI system (which is all GCC) threw a rod.

mathomp4:
@edwardhartnett I have a [reply from one of our devs](https://github.com/GEOS-ESM/NCEP_Shared/issues/17#issuecomment-730704449) which I'll quote here:

> I don't think that there was any difference in the results when compiling with and without -DDYNAMIC_ALLOCATION, in part because the code runs essentially the same if none of the default settings are changed. The dynamic allocation allows increasing the maximum sizes of some of the arrays used in the library, and allowed NCEP to retire their 'supersize' build.
> 
[...]

PR #92, opened 2021-01-08
modified Fortran variable declarations for compatibility with GNU v10+

jbathegit:
Fixes #81 

I modified several internal declarations as discussed in the ticket, and these test successfully on our current platforms, but I have no access to a GNU v10+ compiler to verify them with.   So could I please ask @mathomp4 and @tclune to pull these updates and test them out yourselves on your own systems to verify they're now compatible with that compiler, before I/we merge this new branch into develop?

The changes are just to declarations (no changes to functionality), and the subprograms and code lines involved are already well-covered by the existing test suite, so no new test cases are needed.

PR #95, opened 2021-01-12
activate testing of static memory builds for all 4-byte builds on all platforms

jbathegit:
Part of #77 

@edwardhartnett, @aerorahul, @kgerheiser 

As discussed, for now I wanted to re-activate testing for the static memory builds.  I pushed up this branch to test it on gcc-9 and gcc-10, and I was actually expecting the build to fail for gcc-10 b/c it doesn't yet include the src changes from my jba_mstabs branch.  Basically, I wanted to verify that I could reproduce the compile error that Matthew Thompson reported in #81, and then I was going to merge this branch into jba_mstabs and test it together with the src changes in that branch.

So again, I was expecting this particular push to fail, but not in the way that it actually did.  Specifically, it looks like it didn't fail because of the error noted in #81, but rather because it never created test executables for the static memory builds, and therefore it couldn't find those targets in the "build" step.  See https://github.com/NOAA-EMC/NCEPLIBS-bufr/actions/runs/481158783  I'm not sure why this happened or how to fix it (I'm still a bit of a CMake newbie ;-), so I would appreciate any suggestions on how to fix this.

Thanks very much for any help you can provide here.  I seem to recall a discussion in an earlier issue about problems trying to generate static memory builds with non-Intel compilers, so not sure if this is related to that or if that was ever even resolved(?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,12 +43,8 @@ if(APPLE)
 endif()
 
 # Determine the kinds upfront (both src/ and test/ need them).
-# Both GNU and Intel support DYNAMIC_ALLOCATION
 list(APPEND kinds "4_DA" "8_DA" "d_DA")
-# Only Intel supports STATIC_ALLOCATION
-if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
-  list(APPEND kinds "4" "8" "d")
-endif()
+list(APPEND kinds "4" "8" "d")
 
 # Set common to the package compiler flags
 if(CMAKE_C_COMPILER_ID MATCHES "^(GNU)$")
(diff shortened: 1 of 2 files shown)


jbathegit:
I kept digging on my own, and I found a corresponding line in CMakeLists.txt that also needed to be modified similar to the one I first modified in test/CMakeLists.txt  This now gets past the previous roadblock, and furthermore it now reproduces the same compile error noted by Matthew Thompson.  So as noted earlier, I'll merge these two changes into my jba_mstabs4 branch and we'll see what that does...

jbathegit:
Bingo, that did the trick.  Everything built and tested successfully using the jba_mstabs branch after I merged in the jba_test4 fixes.  So I'll close this PR and we'll revert to the previous one in #92 to merge all of the updates for both new branches back into develop.

PR #96, opened 2021-01-19
updates to Doxygen documentation, including master BUFR tables

jbathegit:
See https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/ for a preview of what the changes would look like when displayed via Doxygen.

Note that the new tables directory (containing master BUFR tables) is automatically copied to the install area upon a successful build of the source code.


src/CMakeLists.txt
```diff
@@ -102,7 +102,14 @@ install(
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
-### Package config
+# Copy the tables directory to the install area if the library builds are successful
+add_custom_command(
+  TARGET ${LIB_TARGETS}
+  POST_BUILD
+  COMMENT "Copying tables directory to install area"
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/tables ${CMAKE_INSTALL_PREFIX}/tables)
```
aerorahul, line 110: I think you want this to be `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`

aerorahul (commented):
Thanks for adding the tables.
Q1. Are the tables used in any of the codes?  I didn't think so, so just asking.  Are these just for documentation then?  If so, could they be in `docs/tables`?
Q2. A comment on the installation of the tables.  When will an end user need these tables from the installation directory?

src/CMakeLists.txt
jbathegit: Thanks @aerohul for taking a look at this!

The master tables are used as input within numerous operational codes on the WCOSS, so they're not just for documentation.  In fact, they're required for any user code that wants to read/decode BUFR messages according to the internal data description within the messages themselves, rather than having to know this information ahead of time and provide it to the software as a separate input file.  These tables aren't yet used by any of the example codes in `utils`, though I'm planning to provide at least one such example code in the future.

However, 4 of the 10 example codes in `test` (specifically, test_IN_1, test_IN_4, test_IN_5 and test_OUT_4) do already use master BUFR table files, though at the moment copies of these tables are being provided separately within the `bufr.tar` archive file that is pulled from emcrzdm, rather than having the codes point to the new `tables` subdirectory.  If you want I could work towards doing that, but the tricky thing is that the directory where these tables live needs to be known to those test codes at compile time.  Maybe if I created .in versions of those test codes and then used CMake variable substitution at build time (Hmm.. ;-)

The reason I used `CMAKE_SOURCE_DIR` instead of `PROJECT_SOURCE_DIR` was simply because it was already being used in src/CMakeLists.txt and docs/CMakeLists.txt, whereas the latter is never referenced in any of the existing CMakeLists.txt files.  So I was just trying to be consistent with what appeared to be current practice.

-

aerorahul: thanks for explaining the tables and their use.  I am obviously not familiar with their use.
You can keep CMAKE_SOURCE_DIR.  That and PROJECT_SOURCE_DIR are one and the same unless one includes nested projects.  I doubt that will ever happen in this library.

aerorahul (approved):

src/CMakeLists.txt
aerorahul, line 110: I should have caught this at review of this PR.
This needs to be fixed with an `install` command.
For testing, we need to link the tables from the source directory to the build directory.

PR #98, opened 2021-01-21
use tables directory for testing, rather than legacy master tables from bufr.tar archive

### LINKED ISSUE #97: use master tables in tables directory for testing, rather than legacy copies of same files from bufr.tar archive
4 of the 10 example codes in test (specifically, test_IN_1, test_IN_4, test_IN_5 and test_OUT_4) use master BUFR table files, which as of today are now part of the distribution package within the new `tables` subdirectory.  Up until now, static copies of some of these same files have been provided separately within the `bufr.tar` archive file that is pulled from emcrzdm.

The goal now will be to have the above test codes point to the master BUFR table files in `tables`, and at which point we can remove the legacy copies from within `bufr.tar`. The tricky piece is that the directory where the master BUFR tables live needs to be known to these test codes at compile time and provided as an argument to subroutine mtinfo.  One idea may be to use CMake variable substitution to supply this argument within the source codes at compile time.

jbathegit:
Fixes #97

With this update, 10 legacy master table files have now been removed from the bufr.tar archive on emcrzdm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,8 @@ if(ENABLE_PYTHON)
   add_subdirectory(python)
 endif()
 
+add_subdirectory(tables)
+
 add_subdirectory(test)
 
 # Generate doxygen documentation if desired.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -102,13 +102,6 @@ install(
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
-# Copy the tables directory to the install area if the library builds are successful
-add_custom_command(
-  TARGET ${LIB_TARGETS}
-  POST_BUILD
-  COMMENT "Copying tables directory to install area"
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/tables ${CMAKE_INSTALL_PREFIX}/tables)
-
 # Package config
 include(CMakePackageConfigHelpers)
 set(CONFIG_INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
--- a/tables/CMakeLists.txt
+++ b/tables/CMakeLists.txt
@@ -1,0 +1,85 @@
+list(APPEND bufr_tables
+  bufrtab.CodeFlag_LOC_0_7_1
+  bufrtab.CodeFlag_STD_0_13
+  bufrtab.CodeFlag_STD_0_14
+  bufrtab.CodeFlag_STD_0_15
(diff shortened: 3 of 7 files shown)


jbathegit:
Hi @aerorahul and @edwardhartnett , could at least one of you please look this over and approve it so I can get it merged?  Thanks!

aerorahul (commented):
What is `../install/tables`?
Does that imply the package needs to be installed before the test can be run?
I don't think using `../install` is a good idea.

aerorahul (commented):
What you could do is to link the tables in the build directory from the project directory via 
`CMakeLists.txt`  like we do for `testfiles` etc.

jbathegit:
No, the testing via `ctest` can still be done prior to the formal install via `make install`.  The way I had set this up was to copy the tables to the install directory if the library builds are successful (see the following lines in src/CMakeLists.txt):

\# Copy the tables directory to the install area if the library builds are successful
add_custom_command(
   TARGET ${LIB_TARGETS}
   POST_BUILD
   COMMENT "Copying tables directory to install area"
   COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/tables ${CMAKE_INSTALL_PREFIX}/tables)

So the tables get copied over to the install directory only after the library is successfully built, but still prior to testing and the formal install.  So I think this is OK.

aerorahul:
> No, the testing via `ctest` can still be done prior to the formal install via `make install`. The way I had set this up was to copy the tables to the install directory if the library builds are successful (see the following lines in src/CMakeLists.txt):
> 
> # Copy the tables directory to the install area if the library builds are successful
> add_custom_command(
> TARGET ${LIB_TARGETS}
> POST_BUILD
> COMMENT "Copying tables directory to install area"
> COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/tables ${CMAKE_INSTALL_PREFIX}/tables)
> 
> So the tables get copied over to the install directory only after the library is successfully built, but still prior to testing and the formal install. So I think this is OK.

Nothing should go to install area before a `make install`.
We need to remove the copying from `POST_BUILD` and implement a proper `install`.

aerorahul:
@jbathegit I pushed a commit to this branch that now installs the tables when `make install` is invoked.  It also symlinks the tables in the build area for testing.
The test files were updated to use the tables from the build area and not the install area.  The install area can and should be anything the user chooses not just `../install`, which is not the appropriate installation prefix.

jbathegit:
Thanks @aerorahul !  I had been working on making the changes you suggested, but you beat me to it ;-)

I especially like the symlink approach during testing, so that the actual files are only copied over during the install.  I wouldn't have thought of that one myself ;-)

Again, thanks for the help and much appreciated!

jbathegit:
So I guess now we need @edwardhartnett to approve this before we can merge it ;-)

aerorahul (approved):
I approve my changes.  😀

PR #100, opened 2021-01-29
more Doxygen updates to documentation

jbathegit:
This may be a lot of files, but it's all documentation (no source code changes ;-)  It also includes Doxygen markdown in several C language codes, which can serve as prototypes for all of us going forward with documentation of other C code in other libraries.


aerorahul (approved):
looks good.

src/crbmg.c
```diff
@@ -1,52 +1,55 @@
...
+ *  @param[out] iret    - integer: return code
+ *                         - 0 = normal return
+ *                         - 1 = overflow of BMG array
+ *                         - 2 = "7777" indicator not found in
+ *                               expected location 
+ *                         - -1 = end-of-file encountered while
+ *                               reading
+ *                         - -2 = I/O error encountered while reading
+ *
+ * <p>This subroutine is designed to be easily callable from
```
edwardhartnett, line 32: I don't think you need the <p> tags.

edwardhartnett:
Great to see all this documentation progress. Are you updating the gh-pages branch as you go?

src/crbmg.c
jbathegit: I've found in some cases that if I don't include the `<p>` tags, then all of the text discussion gets mashed up together in one big block.  So to be safe and to keep the output more readable, I've been including them, but only where I've found from experience that they're needed.

jbathegit:
I haven't continued to update the `gh-pages` beyond my initial update a couple of months ago, because since then there have been a number of source changes that aren't part of v11.3.2 (which is the latest release version), and I didn't want to be documenting anything that was unique to v11.4.0 (which will be the next release version) under the v11.3.2 label.  My plan is to have a new release version ready to go in the next month or two, and certainly at that point I/we will get all of the `gh-pages` documentation sync'ed up.

Issue #48, opened 2020-11-03
transfer text from external HTML docs to user_guide.md and code files in repo


edwardhartnett:
As discussed in meeting on 11/2/20, there is an external HTML documentation site for NCEPLIBS-bufr. It should be referenced in the README. (@jbathegit can you put the link in this issue or in the README?)

This external documentation contains additional text that describes NCEPLIBS-bufr and its functions.

Some of the external docs refers to specific functions and can be moved to those code files. Other parts are general and will be moved to user_guide.md (or perhaps other .md files in the docs directory if needed).

This is related to #15

jbathegit:
The URL for the external HTML documentation is now in the new UserGuide.md, as of PR #76

jbathegit:
This has been completed effective with the v11.5.0 release.

PR #101, opened 2021-02-09
more Doxygen updates to documentation

jbathegit:
Part of #48


src/cwbmg.c
```diff
@@ -13,11 +13,10 @@
...
  *  @date 2005-11-29
  *
- *  @param[in] bmg     - character*(*): array containing BUFR
- *                        message to be written
- *  @param[in] nmb     - integer: size (in bytes) of BUFR message
+ *  @param[in] bmg     - char*: BUFR message to be written
+ *  @param[in] nmb     - f77int: Size (in bytes) of BUFR message
  *                        in bmg
- *  @param[out] iret    - integer: return code
+ *  @param[out] iret    - f77int: return code
```
aerorahul, line 19: what is f77int?

jbathegit: It's a typedef defined in bufrlib.h, equivalent to either an int or a long depending on whether it's a _4 or _8 build.

aerorahul: Thanks!

aerorahul (approved):
lots of dox updates.

edwardhartnett:
Excellent work cleaning up these documentation blocks! Do we have one of our github workflows building the docs?

If not, you might want to add that. Simply make sure that doxygen is apt-get installed, and then add -DENABLE_DOCS=Yes to the cmake command in one of the workflows in .github/workflows...

Issue #91, opened 2021-01-05
test codes for utils programs


jbathegit:
As noted in the discussion for #88 , all sample programs under /utils need to have an associated test code under /test.  Currently, only gettab has an associated test code, so we still need them for readbp, readmp, binv, sinv, and cmpbqm at some point.

There's no specific deadline for this task; rather, this is just a placeholder for tracking this issue and for any further discussion or developments.

Issue #32, opened 2020-10-28
Identify areas for additional testing


edwardhartnett:
@jbathegit run these commands:

```
cmake -DCMAKE_Fortran_FLAGS="-fprofile-arcs -ftest-coverage" -DENABLE_TESTS=ON .. && make clean && make VERBOSE=1 && make test
gcovr -r ../ . --html --html-details -o example-html-details.html
```

Then open the file example-html-details.html to see the code coverage report.

Here's a snapshot: 
![image](https://user-images.githubusercontent.com/38856240/97475565-07bf9080-1913-11eb-8175-79d4dcfc4394.png)

The overall test coverage is 50%. This is pretty good, but we need it to be above 70% for sure. Can you identify areas for additional testing? (Alternatively, some of the code may be dead and can be removed instead of tested, which is even better.)

Some functions that leap out: src/copybf.f, src/copymg.f, src/copysb.f, src/cpbfdx.f, src/cpdxmm.f, src/cpymem.f, src/cpyupd.f are all untested...

As we identify areas for increased testing, add issues. When you have made a complete pass through the coverage report, and identified the first pass of tests that are needed, we can close this issue.

jbathegit:
Where does the gcovr command live?  I don't see it on the WCOSS - do I need to run gcovr on a different system?

Either way, I am working on some new tests as we speak ;-)

edwardhartnett:
gcovr is a Unix utility. It can be installed with:

`sudo apt-get install gcovr`

It's a standard GNU tool so if you ask the sysadmins, they should be able to easily install for you. But then you have to, as you note, tar up the web pages and put them somewhere you can look at them.

I work on a stand-alone Linux box, not on the WCOSS systems.

jbathegit:
Well, there's no way they'll install this on the WCOSS, and even though I have my own stand-alone NCWCP Linux box (emc-lw7-jator), it isn't installed there either at the moment, and as backed up as the sysadmins are right now it could be weeks/months before they could possibly get around to this.  

Is your Linux box on the NCWCP network that I could maybe ssh into using my jator account?  Or, failing that, would you mind just going ahead and creating a gcovr listing for me to look at (a full one, not just the snapshot shown above? ;-)

edwardhartnett:
Sure, no problem. Some day I hope we can have these coverage reports generated on-line automatically every day. But we're not there yet. ;-)

[...]

PR #103, opened 2021-02-18
add debufr to sample utilities, along with corresponding test cases

jbathegit:
Part of #91 
Part of #32

### DIFF
--- a/utils/debufr.f
+++ b/utils/debufr.f
@@ -1,0 +1,461 @@
+C> @file
+C> @brief Fortran-language code for debufr utility.
+
+C> This module is used to share information between
+C> subroutine fdebufr() and subroutine openbt(), since the latter
+C> is not called by the former but rather is called directly
+C> from within the BUFRLIB software.
+
+	MODULE Share_Table_Info
+	    CHARACTER*120	ctbldir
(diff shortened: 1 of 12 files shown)


utils/debufr.f
```diff
@@ -0,0 +1,461 @@
+C> @file
```
aerorahul, line 1: why C and fortran both?

jbathegit: This utility has a lot of UNIX command line options, which at least historically are generally easier and more standard to process in C than in Fortran.  So the driver is written in C, and then most of the remaining code logic is in Fortran.  The library itself has a similar mix of C and Fortran, and as long as the interface across the languages is clean and deliberate, it works and is portable across all platforms.

aerorahul:
I am going to let @edwardhartnett comment on this.

jbathegit:
@edwardhartnett could you please take a look and let me know if you have any questions/concerns about this PR?  If not, I'd like to get it approved and be able to merge it, because I now have another testing-related issue (#104) to work on, and it will likely involve additional updates to some of the same configuration files.

Please let me know - thanks!

edwardhartnett (approved):

Issue #106, opened 2021-02-22
allow encoding of mnemonics which are larger than 32 bits


jbathegit:
A user reported that the software will not allow encoding of Table B elements which are greater than 32 bits.  This was discovered when trying to encode the Table D sequence F5FCWS (3-15-010) for drifting buoy data.  The sequence includes the Table B element SWDE (0-22-069), which has a native bit width of 22 bits but is then increased to 38 bits by the 2-01-144 operator that is also contained within the sequence.

jbathegit:
So it turns out the real solution here isn't to modify the library itself, but rather to just link the application program in question with one of the 8-byte integer builds of the library, because then all integers in internal memory are 64 bits wide, and the encoding works properly.  The user in this case was trying to link to a 4-byte integer build of the library.

But on a related note, I will modify the build of the debufr utility to now link to the libbufr_8_DA.a build of the library rather than the libbufr_4_DA.a, so that it gives proper results in the future if/when it encounters any Table B elements with bit widths greater than 32 bits! ;-)

PR #108, opened 2021-02-24
Std WMO BUFR Support  for C Interface

rmclaren:
This pull request exposes the Fortran function **mtinfo** to make it possible to open a standard WMO bufr file using the c/c++ interface of NCEPLIB-bufr.


edwardhartnett:
Some unit tests?

aerorahul:
> Some unit tests?

Are there unit tests that exercise these routines?
If so, we can wrap those in the c++ calls as tests.

We didn't add new code, just exposed an interface to the underlying fortran routines.

edwardhartnett:
@jbathegit does have some unit tests, but it's not clear what those will do.

What we do for netcdf-fortran (which just calls the existing netcdf-c library) is test each function, but not all combinations of parameters for each function.

In other words, the test is to ensure that the API is working, not to duplicate the tests of the layer below, which will be more extensive.

jbathegit:
There are no "unit tests" per-se in the BUFRLIB, meaning no encapsulated code that just tests the functionality of a single particular subroutine or function.  The reason is because the BUFRLIB doesn't work that way, e.g. any use of the BUFRLIB involves calling multiple subroutines or functions in a particular order to collectively accomplish any practical task of encoding or decoding a file of BUFR messages.

That's how our tests work now in the "test" subdirectory of the project, by calling multiple subroutines or functions in a particular order and then verifying the collective output.  But @edwardhartnett is also correct that we all need to get in the practice of providing accompanying test code whenever we make updates to source code.  So maybe what could be done here is to take one or more of the existing Fortran test codes and make a C version of that code which does the same thing, and could even be compared to the same output to verify the result.  For example, test_IN_4.f calls openbf, readmg, ireadsb, mtinfo, ufbint, ufbrep and closbf, so maybe create a test_IN_4.c which does the same thing but calls the bufr.interface.f90 wrappers for those subprograms instead of calling those subprograms directly?  And when doing so you could also replace the call to readmg with one for the ireadmg wrapper.  That seems like it would be a sufficient test for most/all of the stuff in bufr.interface.f90, unless I'm missing something(?)

rmclaren:
Ok, I added some tests for the c interface. Mostly making sure it doesn't crash.

rmclaren:
test_debufr_1 and 2 are failing.

aerorahul:
> test_debufr_1 and 2 are failing.

debufr requires linking with `bufr_8`. 

When will we be properly overloading functions rather than creating `_4` and `_8` and `_d` libraries?

test/test_c_interface.c
```diff
@@ -0,0 +1,152 @@
+/** @file
```
edwardhartnett, line 1: Nice!

Thanks!

jbathegit:
At the moment, debufr on the develop branch doesn't require linking with `bufr_8`, though it will at some point following a future PR for #106   What happened here is that I was in the middle of testing a new bufr.tar file on the emcrzdm server while in the process of getting ready to submit that PR, and there are a couple of minor whitespace changes in the expected debufr output now.  I just restored the previous copy of bufr.tar, so @rmclaren if you want to re-run the check it should work now.

@aerorahul maybe I'm missing something, but the different `_4`, `_8` and `_d` versions has nothing to do with method overloading, which as I understand it means you can have different call signatures for a single subroutine or function (or method, in Java or C++ speak).  Instead, the different build versions are due to the fact that there are many application programs out there which are compiled using 4-byte integers vs. 8-byte integers, etc. and need to link to libraries that are built the same way.

edwardhartnett:
@aerorahul I would love to address the 4/8/d issue in all the NCEPLIBS. But can it be done in a way that is compatible with existing user code? Or do we have to change all code everywhere? I think we can do it without user code changing but need to test that.

Also, how would be handle a mix? For example if package A goes to overloaded functions, and package B uses package A, if package B still builds 4/8/D versions, will everything still work? I believe it will.

Right now I'm heavily engaged in testing UFS_UTILS...

jbathegit:
@rmclaren thanks for ginning up a test code!  I see where you're verifying the returned values from ufdump_f and ufbrep_f within test_readBufrFile() so that's great.  But I don't see where you're verifying the output in test_dxDump() or test_mtInfo(); rather, it looks like you're just calling dxdump_f and mtinfo_f respectively, but not actually verifying that they produce the expected results.  Am I missing something here?  Either way, this is definitely good progress, and thanks again!

@edwardhartnett OK maybe I'm the one who's confused here, but I don't understand in theory how we could have a single build of the software that works for all user codes.  You've probably got more experience in this area than I do, and if there's really a way to do that, then I'm all for it.  But off the top of my head I don't see how that's possible(?)

jbathegit:
@rmclaren yes I agree it's a pain to have to make a trivial tweak and new commit just to re-run the integration test, but at least in my experience I haven't figured out any other way to re-run the test on a previous commit.

test/test_c_interface.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+
+void test_dxDump()
+{
+    open_f(BUFR_FILE_UNIT, INPUT_FILE);
+    open_f(DUMP_FILE_UNIT, TABLE_DUMP_FILE);
+    openbf_f(BUFR_FILE_UNIT, "IN", BUFR_FILE_UNIT);
+
+    dxdump_f(BUFR_FILE_UNIT, DUMP_FILE_UNIT);
```
jbathegit, line 126: How do you know dxdump_f gave the correct results?

test/test_c_interface.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+
+void test_mtInfo()
+{
+    open_f(BUFR_FILE_UNIT, INPUT_FILE);
+    openbf_f(BUFR_FILE_UNIT, "IN", BUFR_FILE_UNIT);
+
+    mtinfo_f("testfiles", TABLE_1_FILE_UNIT, TABLE_2_FILE_UNIT);
+
```
jbathegit, line 140: How do you know mtinfo_f did what it was supposed to?  You aren't really using these master tables for anything in this test.

aerorahul:
Go to Actions and hit re-run previous action.

rmclaren:
@jbathegit Thats right, I just call them and see if thy crash or not. Are there any bufr files in testinput that are standard WMO bufr files (for mtinfo). Not sure how I would verify the output for dxdump very easily... I'm almost thinking it would just be easier to get rid of it at this point... deosn't really add anything to this interface as it can only be used to generate a file. Easier just to call the gettab utility.

test/test_c_interface.c
rmclaren: I could either remove the test, or just remove dxdump from the interface entirely... Its not really adding anything at this point (there really isn't a reason to need to call it from c).

test/test_c_interface.c
rmclaren: Wondering if you have any Standard WMO BUFR files in the testfiles/data directory, otherwise I could add one Greg sent me.

jbathegit:
@aerohul thanks for the tip!  I'll try to remember that ;-)

@rmclaren I presume someone created the dxdump_f binding for a reason.  Basically, what you could do is run a diff or a cmp on the generated output file and make sure it matches a static file in testfiles/testoutput.  But if if that binding has been there and you weren't the one who added it, then I guess I can't ask you to now be responsible for adding test code for it ;-)  Just FYI that this is why from now on we're always asking for test code corresponding to any source changes.  And on that note, for the new mtinfo_f binding, I presume you have some C application that you want to use this in?  Is there any way a simple test code could be generated from that application?  Or another alternative as I mentioned earlier is to make a C version of test_IN_4 which does the same thing as the existing Fortran version, and could even be compared to the same output to verify the result.  The reason I suggested test_IN_4 is because it's an example of a code that already calls mtinfo, as well as many of the other Fortran routines for which there are now C bindings in bufr_interface_f90

test/test_c_interface.c
jbathegit: The tables subdirectory under the main project is chock-full of master table (i.e. Standard WMO BUFR) files that can be used with mtinfo_f.  This subdirectory already gets copied over to the build area for testing, so you can reference it there from any test code.  But the issue here is that mtinfo doesn't really do anything other than set the directory and unit numbers for reading master tables, which aren't actually done until a later call to ireadmg.  So until you call ireadmg, there's no way to know that it really found the tables, and which is why mtinfo can really only be tested within the scope of a larger program which is doing other things as well.  Again, test/test_IN_4.f shows a good example of this.

jbathegit: @aerorahul the current test doesn't really ensure the file is present; again, all mtinfo does is just set the directory and Fortran logical unit numbers.

I just made a clone of this branch and will see if I can cook up a quick C test.  This would be good to have anyway.  But I'm presuming @rmclaren also has some C application he wants to use mtinfo with, and which is why he added this C binding to the bufr.interface.f90 file in the first place(?)

test/test_c_interface.c
```diff
@@ -0,0 +1,152 @@
...
+    @brief This file is a quick test of bufr_c_interface_mod.
+*/
+
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "bufr.interface.h"
```
jbathegit, line 13: How does the bufr.interface.h header file get located during compilation, since that header file is in the src subdirectory, but this test code (test_c_interface.c) is in the test subdirectory?  Are you using the -I option when you compile the test code?  If so, I didn't see where that was specified.

Again, just trying to set up my own test code here ;-)

test/test_c_interface.c
aerorahul: then the test is to just set the directory and Fortran logical numbers.
It should do nothing more, because mtinfo does nothing more.

src/bufr.interface.f90
```diff
@@ -1,3 +1,12 @@
+!> @file
```
edwardhartnett, line 1: It's not usual to have an extra "." in the filename. Can this file be named bufr_c_interface.f90, to accord with usual conventions?

src/bufr.interface.f90
```diff
@@ -7,15 +16,23 @@ module bufr_c_interface_mod
...
 
 contains
 
 !Private
+
+!> @brief This function turns a c string into a fortran string.
+!>
+!> @param[in] c_str - c_char: pointer to a \0 (null) terminated c string
+!> @param[out] f_str - character(:): fortran string
+!>
```
edwardhartnett, line 35: Please put an author tag on every function documentation block.

src/bufr.interface.f90
```diff
@@ -77,13 +100,17 @@ subroutine closbf_c(bufr_unit) bind(C, name='closbf_f')
 end subroutine closbf_c
 
 
+subroutine exitbufr_c() bind(C, name='exitbufr_f')
```
edwardhartnett, line 103: All subroutines and functions need doxygen documentation.

src/bufr.interface.h
```diff
@@ -1,21 +1,42 @@
-/** @file */
```
edwardhartnett, line 1: No "." in files names, except for the file extension please.

test/test_c_interface.c
aerorahul: `target_link_libraries` when linking with the library will include the path to the header file.
The header file is installed when the library is installed.

test/test_c_interface.c
edwardhartnett: Remove it from the interface if it should not be there.

rmclaren: @jbathegit I had actually originally added this as sort of an excercise when I first got started wrapping fortran in C. Actually I added everyhtin in that file. I think I'll take @edwardhartnett advice.

test/test_c_interface.c
rmclaren: I was going to add a new test readStdWmoBufrFile that calls mtinfo (similar to readNcepBufrFile). 

@jbathegit The function is being used in a c++ application but not in a form where you could just copy and paste the code into a test application...

jbathegit:
I just pushed up an additional commit which includes:

1. a new test_c_interface_2.c code which fully tests the mtinfo_f function, and corresponding updates to CMakeLists.txt
2. a small tweak to the existing test_c_interface.c code to get it to work on Intel
3. a rename of bufr.interface.[h|f90] to bufr_interface.[h|f90]

@rmclaren or @aerorahul, do you know who the original author of the bufr.interface.[h|f90] codes was?  If it was one of you, could you please add another commit with some Doxygen documentation, as noted by @edwardhartnett? (Thanks!)

jbathegit:
Crap, the new test code just failed on gcc10, but it works fine on gcc9 and on Intel.  Does anyone know where there's a machine with a gcc10 compiler that I could test with?  Or if one of you have access to such a machine would you mind taking a look at my test_c_interface_2.c code and see what it may be squawking about?  The only gnu compiler I have access to on the WCOSS or on my NCEP workstation is a really old one (i.e. gnu v4.8.5)

Thanks for any help!

aerorahul:
> Crap, the new test code just failed on gcc10, but it works fine on gcc9 and on Intel. Does anyone know where there's a machine with a gcc10 compiler that I could test with? Or if one of you have access to such a machine would you mind taking a look at my test_c_interface_2.c code and see what it may be squawking about? The only gnu compiler I have access to on the WCOSS or on my NCEP workstation is a really old one (i.e. gnu v4.8.5)
> 
> Thanks for any help!

The action runs gcc-10.  You can edit the action yaml file and put in what you wish and re-trigger the action.  You could even just recreate the action if you would like.
I don't have a machine, but I do have a container on my mac that has a gcc-10 setup.

jbathegit:
> The action runs gcc-10. You can edit the action yaml file and put in what you wish and re-trigger the action. You could even just recreate the action if you would like.
I don't have a machine, but I do have a container on my mac that has a gcc-10 setup.

But running it via the re-trigger wouldn't show any print output (i.e. stdout), so how would I be able to tell where in the code it's failing?  I think(?) what is needed is to actually build the executable and run it manually in a gcc-10 environment so we can see exactly what it's doing, maybe using gdb (i.e. the gnu debugger).  Otherwise, I'm not sure how to really troubleshoot this.

Or maybe I'm missing something here?

rmclaren:
I added the requested doc strings to the .f90 and .h files. There is a lot of duplicate docs between the two of them, however. I'd like to suggest we should remove the doc strings on those functions in the .f90 file that are being bound as C style functions via cbind as these functions can not be called directly (so doc strings are of limited value).

rmclaren:
@jbathegit Why not just add your test to the test_c_interface.c file instead of creating a whole new one?

edwardhartnett:
One answer is that you get an ubuntu machine from NOAA. Is that possible and easy?

aerorahul:
> One answer is that you get an ubuntu machine from NOAA. Is that possible and easy?



> One answer is that you get an ubuntu machine from NOAA. Is that possible and easy?

Hahahaha

rmclaren:
Ok, there is something weird going on here. Changing the the length of the subset string causes it to crash reliably everywhere. Bringing up the code in my debugger I see the following:

image

I'll take a closer look at this.

edwardhartnett:
Have you tried running with the address sanitizer or valgrind?

rmclaren:
I pretty sure there is a problem with how I'm copying the fortran string to the c string. I've been meaning to improve this anyways. So I'll let you know when I check in changes.

jbathegit:
@rmclaren, SUBSET_STRING_LEN = 8 definitely won't work, because the value that should come back in msg_subset is "MSTTB001" which is 8 characters, plus you need an additional space for a trailing null.  So that macro needs to have a minimum value of 9, unless I'm missing something.

jbathegit:
> > One answer is that you get an ubuntu machine from NOAA. Is that possible and easy?
> 
> > One answer is that you get an ubuntu machine from NOAA. Is that possible and easy?
> 
> Hahahaha

I have to second Rahul's reply here.  We might as well ask them for a keg of beer while we're at it ;-)

rmclaren:
@jbathegit  Depends on whether you treat is as a "c string" or as a character array... I did do something really  stupid with respect to this in my string copying function, but actually there is a bigger issue.

When you call NCEPLIB-bufr::readmg as follows:

```fortran
character(len=25) :: f_subset     !! 25 cause I figured that was long enough to hold any subset name
ires = ireadmg(bufr_unit, f_subset, iddate)
```
You get back the following in (in the case where it fails):
MSTTB001\020\271\265\362\271U\000\000\000\000\000\000\000\000\000\000P

The way strcmp works is it compares strings until it hits a non-ascii character (at least for me, officially it runs until hitting the end of the array or a null character). The reason it fails is that \020 is a valid ascii character... In short you get back the subset and then some garbage (guessing whatever is next in memory). Sometimes that garbage includes valid ascii characters in the next position... so strcmp reports the strings are not equal.

I've tried using len_trim, len, trim to get back just the subset string, but none of it was much help... I'm not sure how you go about safely retrieving the actual subset string (without the unwanted stuff)...

jbathegit:
How about using strncmp instead of strcmp?  No subset ID should ever be longer than 8 characters, so
`if (strncmp(subset, msg_subset, 8))`
might do the trick(?)

rmclaren:
That would solve the immediate problem, yes.

edwardhartnett:
You need to explicitly set the null value in the C string, if you are copying from a fortran string.

rmclaren:
@edwardhartnett I'm gonna treat it as a character array, not a string... setting null at the end is pointless in this case as it is impossible to determine where the string actually ends.

rmclaren:
Ok, added it back.

jbathegit:
Looks good to me.  Thanks to all of you for your help with this!

jbathegit (approved):

aerorahul:
Please consider a release or tag at this point since this PR changed the access to the C++ api by replacing `bufr.interface.h` with `bufr_interface.h`
Our downstream applications are using these C++ interfaces.

rmclaren:
Not too late to revert the name change... just saying.

PR #112, opened 2021-03-03
new and updated tests for 8-byte builds of the library

### LINKED ISSUE #106: allow encoding of mnemonics which are larger than 32 bits

jbathegit:
Fixes #106


test/CMakeLists.txt
```diff
@@ -111,15 +115,16 @@ function(bufr_add_test_preAPX TESTNAME EXENAME)
            COMMAND ${CMAKE_BINARY_DIR}/bin/test_wrapper_OUT.sh ${EXENAME} "Y")
 endfunction()
 
-# test all "4" (4 byte real, 4 byte int) DYNAMIC_ALLOCATION builds
-list(APPEND test_kinds "4_DA")
-# test all "4" (4 byte real, 4 byte int) STATIC_ALLOCATION builds, except OUT_3 and OUT_4 as noted below
-list(APPEND test_kinds "4")
+list(APPEND test_kinds_4 "4_DA")
+list(APPEND test_kinds_4 "4")
```
aerorahul, line 119: could be condensed into `list(APPEND test_kinds_4 "4_DA" "4")`

test/CMakeLists.txt
```diff
@@ -161,10 +166,10 @@ foreach(test_src ${test_OUT_2_srcs})
   endforeach()
 endforeach()
 
-# OUT_3 and OUT_4 tests
+# OUT_3 and OUT_4 tests; these are for "DA" (dynamic allocation) builds only
 foreach(test_src IN ITEMS ${test_OUT_3_srcs} ${test_OUT_4_srcs})
   string(REPLACE ".f" "" testPref ${test_src})
-  foreach(kind ${test_kinds})
+  foreach(kind ${test_kinds_4})
```
aerorahul, line 172: the if statement on the next line will never have 8 or d, so please update the logic

aerorahul (commented):
most changes look good.  a couple of suggestions.

edwardhartnett (approved):

test/CMakeLists.txt
jbathegit: Noted

test/CMakeLists.txt
jbathegit: True, but in the future I may add some additional tests of 8 and d builds, so I'd prefer to leave that line as is for now.

PR #114, opened 2021-03-05
more testing and more documentation

jbathegit:
Part of #32

### DIFF
--- a/src/getvalnb.f
+++ b/src/getvalnb.f
@@ -1,56 +1,57 @@
 C> @file
-C> @author J @date 2012-09-12
-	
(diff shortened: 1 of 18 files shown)


aerorahul (approved):
I don't mind approving, but we should try and have separate PRs for documentation, code and tests. 
It is difficult to review big PRs.

jbathegit:
I understand your point, and I agree it can be difficult and unwieldy for reviewers when too many changes are bunched together in a single PR.  But in this case, and other than correcting a few typos, the documentation updates were to add Doxygen docblocks for the exact same routines that were being added to the testing.  In other words, while I was adding tests to the test codes, I went ahead and included the Doxygen updates for those same routines as well.

PR #119, opened 2021-03-09
consolidate github actions in one file

aerorahul:
Following up from a discussion with @jbathegit in another thread, this PR consolidated the actions into a single yaml file.


jbathegit:
Thanks for this!  It looks good to me, and I'll approve and merge it if/when the final CI tests ever finish.

jbathegit:
Actually, it looks like we now have some additional checks that weren't there before (there were 12 previously, now 16).  Was this your intent @aerorahul?

aerorahul:
> Actually, it looks like we now have some additional checks that weren't there before (there were 12 previously, now 16). Was this your intent @aerorahul?

hmm .. what checks are missing?
I added all viable combinations of `macOS`, `ubuntu`, `clang`, `GNU-9`, and `GNU-10`.
There should be 12 total checks.
macOS + clang + gfortran-9
macOS + clang + gfortran-10
macOS + gcc-9 + gfortran-9
macOS + gcc-10 + gfortran-10
ubuntu + gcc-9 + gfortran-9
ubuntu + gcc-10 + gfortran-10

What am I missing?

We shouldn't be mixing gcc-9 with gfortran-10 and gcc-10 with gfortran-9 anyways.

jbathegit:
Your list looks correct, meaning there should be 12 total checks since each one is done on both push and on PR.  But it looks like it's trying to run 16 checks, because the 12 expected ones have already completed and 4 appear to be hung.

aerorahul:
I don't know what does expected mean here.
Can you approve and see if we can merge based on the passing of the current list of checks which seem to be correct.

PR #122, opened 2021-03-15
Expose tables directory to the user 

aerorahul:
Add a variable in package config so downstream dependencies can find it.

### DIFF
--- a/cmake/PackageConfig.cmake.in
+++ b/cmake/PackageConfig.cmake.in
@@ -15,7 +15,7 @@ get_target_property(@PROJECT_NAME@_BUILD_TYPES @PROJECT_NAME@::@PROJECT_NAME@_4_
 
 check_required_components("@PROJECT_NAME@")
 
-set(bufr_TABLES_DIR "${PACKAGE_PREFIX_DIR}/tables")
+set(@PROJECT_NAME@_TABLES_DIR "${PACKAGE_PREFIX_DIR}/tables")
 
 get_target_property(location @PROJECT_NAME@::@PROJECT_NAME@_4_DA LOCATION)
 message(STATUS "Found @PROJECT_NAME@: ${location} (found version \"@PROJECT_VERSION@\")")
--- a/cmake/PackageConfig.cmake.in
+++ b/cmake/PackageConfig.cmake.in
@@ -15,5 +15,7 @@ get_target_property(@PROJECT_NAME@_BUILD_TYPES @PROJECT_NAME@::@PROJECT_NAME@_4_
 
 check_required_components("@PROJECT_NAME@")
 
+set(bufr_TABLES_DIR "${PACKAGE_PREFIX_DIR}/tables")
+
 get_target_property(location @PROJECT_NAME@::@PROJECT_NAME@_4_DA LOCATION)
 message(STATUS "Found @PROJECT_NAME@: ${location} (found version \"@PROJECT_VERSION@\")")


aerorahul:
Clearly the failures are not a result of the changes in this PR, but something else entirely.
Give this sometime, may be someone will fix the issue at the Github end of things.

aerorahul:
as I suspected, the issue was at Github end.  Should be ready to review.

jbathegit:
Where does PACKAGE_PREFIX_DIR get defined?  Is it an intrinsic/pre-set variable in CMake?  Asking b/c I couldn't find a reference to it anywhere in the package, nor in the online CMake reference manual for that matter.  Is it related to CMAKE_INSTALL_PREFIX, and if so how?

Assuming this works, is the idea that we can then use @bufr_TABLES_DIR@ as a CMake macro throughout the rest of the package?  How would regular users access this value?

BTW, would prefer BUFR_TABLES_DIR instead of bufr_TABLES_DIR, since BUFR is an acronym.  An even-more-appropriate name for this value would be MASTER_TABLES_DIR.

aerorahul:
`PACKAGE_PREFIX_DIR` is defined by CMake upon install.  You won't see its references in the source code.
Do a `make install` and then inspect `lib/cmake/bufr-config.cmake` in the installation directory.

This will define a variable `bufr_TABLES_DIR` and set it to the path where the tables are installed.
It is used in a third party library/application when they use the bufr library via:
```
find_package(bufr)
```

The package is named `bufr` in the top-level `CMakeLists.txt`.  You should name it `BUFR`.  This will have a profound impact in a lot of places downstream.

jbathegit:
Fair enough - the bufr vs. BUFR is just a preference on my part, and I can live with the former.  But as for `find_package(bufr)`, and unless I'm missing something, this would only work if users are building their downstream applications using CMake.

At least on the WCOSS, not many folks currently use CMake, which is why I previously suggested to use a CMake macro to hardwire this table directory as the default location within the library itself.  It could still be overridden via a call to mtinfo() if needed, and I think I could still do this using `bufr_TABLES_DIR` as a CMake macro along with configure_make where needed.  In other words, I think I could still do this by using your update as a base and building upon it.

jbathegit:
Actually, now that I think of it, this may not work for my idea if PACKAGE_PREFIX_DIR isn't defined until install time, because `make install` is normally run after testing via `ctest`, and what I have in mind would require the tables to be available during testing.

aerorahul:
> Fair enough - the bufr vs. BUFR is just a preference on my part, and I can live with the former. But as for `find_package(bufr)`, and unless I'm missing something, this would only work if users are building their downstream applications using CMake.
> 
> At least on the WCOSS, not many folks currently use CMake, which is why I previously suggested to use a CMake macro to hardwire this table directory as the default location within the library itself. It could still be overridden via a call to mtinfo() if needed, and I think I could still do this using `bufr_TABLES_DIR` as a CMake macro along with configure_make where needed. In other words, I think I could still do this by using your update as a base and building upon it.

I don't think putting a macro in the source code will work, since this is not replaced until `make install` is called.
Besides, you don't want to put a hard-wired path in the source code at compile time.
Best, I can suggest, if you do want to use a variable is to get it from environment.
If you have an environment variable `bufr_TABLES_DIR` that your code can work with, we will get every bit of functionality as well as relocatable tables.

aerorahul:
@jbathegit 
You mention that on WCOSS, not many people use cmake.  How do they build the bufr library without cmake?  There are no GNU Makefiles here to build the library.
If they just use BUFR library pre-built on WCOSS, can't they just define an environment variable `bufr_TABLES_DIR` and point it to the installation path to read tables or whatever they need from there.

aerorahul:
I don't see anywhere in the library that this path is set.
The subroutine `mtinfo.f` takes an input argument to the path and sets it for the library. 

The test files `test_IN_1,4,5.f` and `test_OUT_4.f` have hard-wired paths `../tables`.  I would make these tests get the path from environment variable e.g. `call getenv("bufr_TABLES_DIR", dir_table)` and then `call mtinfo(dir_table, 90, 91)` etc.

Having hard wired paths should be discouraged, IMO.

aerorahul:
I did not see the hard-wired path in `bfrini.f`. 
Hard-wired paths in a library.
I have to take a break to digest that.

jbathegit:
I'm open for suggestions, but MTDIR needs to have some default internal value, in case the application code never calls mtinfo()

jbathegit:
@aerohul, I'm OK with going ahead and approving and merging this since we're now addressing my question separately in issue #123.

However, and just to clarify my own understanding here, what this change does is allow a user to now have access to a new environment variable `bufr_TABLES_DIR` if/whenever they use CMake to build a separate project, by specifying `find_package(bufr)` when doing so, correct?

aerorahul:
`bufr_TABLES_DIR` is not an environment variable.
It is a CMake variable.

When a downstream package wants to use the bufr library, they do:
```
find_package(bufr)
```

Doing so will set the above variable in the application CMake cache.

We have a C++ API in our JEDI application to open BUFR files. It is not just a call to `openbf`.  The API calls `openbf`followed by a call to `mtinfo`.  The `default` path is the one in the CMake variable, but also allows a user to specify any path desired to point to "tables".

jbathegit (approved):

PR #124, opened 2021-03-19
more documentation, including for all modv files

jbathegit:
I know this is a lot of files, but it's all documentation enhancements, and more than half of the changed files are modv files which only define a single global parameter and thus are very short in length.

What's particularly neat about this commit is that I figured out how to implement the Doxygen @var tag, which creates a global reference to a particular variable name and also generates very nicely-formatted tables for modules.  For examples of what this looks like in Doxygen, see my sample web site at https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/index.html, then click on the "Files" tab, then click on some of the changed files such as cfe.h, moda_bitmaps.F, moda_tababd.F, or any of the modv_*.F files.  All variables defined by these @var tags also end up getting added to a master global catalog, which you can see if you go to the "Main Page" tab, then click on "Data Types List", then click on "Data Fields", then click either "All" or "Variables".

Let me know if any questions - thanks!


jbathegit:
BTW, in all of the modv_*.F files, I also swapped the order of the STATIC and DYNAMIC allocation declarations.  I did this because Doxygen only displays the last one, and I wanted the DYNAMIC declarations to be the ones that get displayed on the web site.  But there are no actual changes to functionality.

src/cfe.h
```diff
@@ -1,24 +1,52 @@
...
+ *  @author J. Ator
+ *  @date 2017-11-16
  */
 
 #define MAX_MEANING_LEN 150
 
 struct code_flag_entry {
     f77int iffxyn;
-	/* Bitwise representation of FXY number to which this entry belongs. */
+    /** @var iffxyn
```
edwardhartnett, line 18: I think you can do this same thing, without the var tag, using a special start for the comment. Like this:

`f77int iffxyn; /*< Bitwise representation of FXY number to which this entry belongs. */`

src/moda_bitbuf.F
```diff
@@ -12,6 +19,16 @@ MODULE MODA_BITBUF
 	  USE MODV_NFILES
 #endif
 
+C>        @var maxbyt
```
edwardhartnett, line 22: I think you can also do this without the var tag, see https://github.com/NOAA-EMC/UFS_UTILS/blob/develop/sorc/chgres_cube.fd/atmosphere.F90 for a good example.

src/cfe.h
jbathegit: You may be right, but I actually prefer the @var tag because it's more descriptive than using a special "<" character.

edwardhartnett:
OK, for static vs. dynamic, what is happening is that whatever macro is applied in the build that is running doxygen is used.

However it's also clear that we need to eliminate what you are calling the static build. It's archaic and adds needless complexity.

jbathegit:
Yes, we've had this discussion before, and we still have an open issue for this in #77.  As noted there, I do agree that we should be able to eliminate the static memory builds as part of a future release of the library.  My plan is to do this not in the next 11.4.0 release, but rather in the next one after that.

jbathegit:
@edwardhartnett do you have any other questions/concerns about this PR?  If not, could you please approve it for merging? (Thanks! ;-)

edwardhartnett (approved):

PR #126, opened 2021-03-19
add new IN_6 test

jbathegit:
Part of #32 

Adding a new test code for several routines which don't yet have any test coverage.  One of them ended up needing a tweak to work correctly for dynamic memory builds, so while I was fixing that source file anyway, I went ahead and added Doxygen updates to that same file as well.  I know it's not best practice to mix testing with source and documentation changes, but since they're all tied together via the same source file, and since this is a small commit to begin with, I figured it would be OK.


jbathegit:
Basically, for the source change, I had to add an #ifdef DYNAMIC_ALLOCATION block to datebf.f, which in turn meant that I had to rename it as datebf.F

aerorahul (approved):
looks good.  Just one comment below in the test.

test/test_IN_6.f
```diff
@@ -0,0 +1,66 @@
...
+
+        CALL UFBMNS ( 18364, cmgtag, imgdt )
+
+        IF ( ( cmgtag .eq. 'NC002003' ) .and.
+     +       ( imgdt .eq. 21031900 ) .and.
+     +       ( NMSUB(iunt2) .eq. 2 ) ) THEN
+            print *, '        UFBMNS -> OK'
+            print *, '         NMSUB -> OK'
+        ELSE
+            print *, '        UFBMNS -> FAILED!!'
```
aerorahul, line 61: since this is a test and it failed, shouldn't we call `BORT` here?
Print will just finish printing and the execution is a success, which is a bit different than a failure.
But, may be its just me and my understanding of the tests that I have written or seen elsewhere.

jack-woollen: I also think the test should have a hard exit if it fails, but I don't know what Jeff has in mind.

jack-woollen (approved):

jbathegit:
Thanks for the reviews.  I wrote these IN test codes with print statements to allow them to be easily run interactively.  That said, within the ctest step of automated builds during CI, the codes all run inside of the test_scripts/test_wrapper_IN.sh script, which includes the following, and where the ${1} argument is the test code executable:

cmd=./${1}

exit \`${cmd} | grep -ic FAILED\`

So the wrapper script will indeed return a non-zero return code if anything fails, which in turn means the ctest will also fail in such cases.

aerorahul:
@jbathegit 
I think using a script is a workaround and flawed.

The external scripts wrapping the tests are generally used when one needs to compare the output produced by the test to a reference, and cannot be adequately tested without making the test overtly complicated.
Where possible, one would use `assert` statements within the test itself to determine if the test is passing or not.  `AAA` is a golden rule of unit testing "Arrange, Act, Assert".

If all the `test_IN` tests are relying on a `grep` from an external shell script.  What if there is a bug in the shell script (granted this is a simple one).  I suggest we remove those and fail within the test execution itself.

If a test fails and you want to see why,  one can re-run the failed tests with verbose logging `ctest --rerun-failed -VV`.
Instead of `print, FAILED`, we should write meaningful abort statements saying what component exactly was tested and what component of the test failed.

PR #129, opened 2021-03-30
remove bufrlib.inc from the library

### LINKED ISSUE #43: should .inc files be included in cmake build lists of files?
@aerorahul I would welcome input on this one.

In autotools systems I would expect to include the .inc files in the list of code files. In that way dependencies can be tracked. (That is, changes in the inc file could trigger rebuild of files that depend on it.

Is that not the case is cmake?

We have a file src/bufrlib.inc (formerly src/bufrlib.prm), but it is not part of the cmake list of sources for the NCEPLIBS-bufr library.

### LINKED ISSUE #123: set default location of master BUFR tables during CMake
Currently, the location of the master BUFR tables is hardcoded to '/gpfs/dell1/nco/ops/nwprod/decoders/decod_shared/fix' in subroutine bfrini.  This path is where these tables live on the WCOSS, and which is why anyone who wants to point to them elsewhere needs to use subroutine mtinfo to do so.  But what if we instead replaced that string with a macro, e.g. '@MASTER_TABLE_DIR@' and then renamed bfrini.f as bfrini.f.in, then used configure_file during CMake to replace the macro with ${CMAKE_INSTALL_PREFIX}/tables?  That way, whatever gets specified as CMAKE_INSTALL_PREFIX when running cmake during the build process would become the default master tables directory for any and all users of that build, and those users would only need to call subroutine mtinfo if they wanted to point somewhere else.

jbathegit:
This commit fully removes the Fortran bufrlib.inc global include file from the library.

Closes #123
Closes #43 

Once this PR is done, I'm putting source changes aside and going back to Doxygen updates for a while! ;-)


jbathegit:
Hi @aerorahul and @jack-woollen, would one of you mind reviewing my PR and let me know if you have any questions or concerns?  Thanks!

src/getbmiss.f
```diff
@@ -23,7 +23,7 @@
 C>
       REAL*8 FUNCTION GETBMISS()
 
-      INCLUDE 'bufrlib.inc'
+      USE MODV_BMISS
```
aerorahul, line 26: I know we relax to doing `USE MODULE_NAME`.
But, I would recommend changing that habit to:
`USE MODULE_NAME, ONLY: USED_ATTRIBUTE_FROM_THE_MODULE`

edwardhartnett: Also, should the filenames be changed to *.F90? Now that you are using F90 language features, .f files will confuse some tools (doxygen?) because the expectation is that .f files contain F77 code and .F90 files contain F90 code...

src/modv_MXS.F
```diff
@@ -0,0 +1,16 @@
...
+C> This module declares and initializes the MXS variable.
+C>
+C> @author J. Ator
+C> @date 2021-03-30
+
+	MODULE MODV_MXS
+
+C>        @var mxs
+C>        Maximum number of entries in the internal string cache
+
```
aerorahul, line 13: Please define private and public and only make mxs public.
This should be a `modv_MXS.f90`
There is no preprocessor involved here.

src/getbmiss.f
jbathegit: I never went through the entire code and changed everything to f90; rather, I had only changed the bfrini.f routine to bfrini.f90 during the previous PR, so that I didn't have to worry about the value of MASTER_TABLE_DIR being longer than 72 characters.

I agree that it would be best to use f90 going forward, and I can start the habit of doing that for any new source files (and I'll also read up some more on ONLY: USED_ATTRIBUTE_FROM_THE_MODULE, which I'd never seen before ;-)  But it would be a major undertaking to modify more than 300 existing source files to change all comments, continuation line punctuation, etc. to use f90 syntax.  For now, all current .f files are indeed F77, and all current .f90 files are indeed F90, so everything is well-defined to Doxygen.

FWIW, I really do appreciate all of the bits of help and advice to move this library forward into a more modern software development and maintenance paradigm.  I've really learned a lot from all of you re: CMake, Doxygen, interfacing C and Fortran, etc. and these will serve me well going forward.  But I don't see a lot of bang for the buck in going through all of the existing code and converting it to f90 just for the sake of f77 vs. f90.  If there's a reason to do so for improved functionality (such as was the case for bfrini.f) then that's one thing, but again I don't see much reason at this point to do so for the entire library.

aerorahul: I meant to suggest using the `use module, only: attribute` in files that are touched going forward.
There is no need at this time to go through 300+ files to do that in one go, so that in a few months some or many will have been updated and make the move easier.  That's just my opinion.

src/modv_MXS.F
jbathegit: OK, will do.  And there are a couple of other existing modf files which also don't have preprocessor directives, so I'll update those as well to keep things consistent (they're very small files), so that only modv files which really need to be preprocessed have the .F extension.

Going forward, I will also try to get in the habit of explicitly specifying PRIVATE vs. PUBLIC - good point!

jbathegit: BTW, since MXS is already declared as a parameter and therefore (by definition) can't be changed by the outside code, then it would be more appropriate here to use `public, protected` here, rather than just `public`, correct?

jbathegit: Never mind, it turns out `parameter` and `protected` are conflicting attributes (you can either have one or the other but not both).  So will just stick with `parameter` and `public` ;-)

jbathegit:
OK, I just pushed up another commit which  (I believe ;-) addresses everything we discussed.  Please let me know if you have any other comments or concerns - thanks!

jbathegit:
mac-os checks are very slow to queue and run, for some reason

aerorahul (approved):
I didn't go with a toothcomb, but did a quick pass.
All seems ok.

aerorahul:
> mac-os checks are very slow to queue and run, for some reason

That is a known issue with Github hosted macOS runners.  CI is not GitHub's "primary" product.  People generally tend to look elsewhere for CI CD tools e.g. TravisCI, CircleCI

kgerheiser:
There is no VM for macOS, so they literally have blocks of Mac Minis stacked in a room somewhere, unlike Linux where they can spin up a new VM on demand.

Issue #131, opened 2021-04-06
Turn 11.4.0 into full release...


edwardhartnett:
It was released as a draft, but it has been used a little, so let's just go ahead and make it a release. That means:

* Release notes
* Do a github release (not just a tag, though this includes a tag)
* Documentation release to gh-pages branch to go with version
* Tarball contains everything needed to build and test (check for cmake files)

@jbathegit I would suggest that you assign this issue to yourself. The Code Manager is the best one to be making releases...

jbathegit:
Will do, though I may need a little hand-holding along the way since this will be my first release from GitHub.  Just want to make sure I don't miss anything important ;-)

edwardhartnett:
I think all you have to do is publish the release and update the docs...

edwardhartnett:
OK, we did this.

PR #132, opened 2021-04-09
more Doxygen updates

jbathegit:
Yes, it's a lot of files, but everything is documentation updates **except** for moda_ifopbf.F being renamed and updated to modv_IFOPBF.f90

I'm going to continue to plug away at a little more documentation, and then by the middle of next week I expect to be ready to do a release as discussed in #131 and in https://github.com/NOAA-EMC/hpc-stack/issues/199


src/ardllocc.c
```diff
@@ -1,42 +1,28 @@
 /** @file
-    @author ATOR @date 2014-12-04
```
edwardhartnett, line 2: I would perfer if you left the author tag at all levels - file, subprogram, and module. Each are different (in general).

I realize in your code you have one subprogram per file, but it needed stay that way. You could will the lottery, move to Tahiti, and then your replacement might decide to merge many subprograms in one file. The file authorship will then be different.

In general, as software libraries like this are maintained over the years and decades, this author information will no longer seem so redundant. Code arrangement drifts over time...

src/ardllocf.F
```diff
@@ -1,40 +1,18 @@
...
-C>   SUBROUTINE SHOULD ONLY BE CALLED DIRECTLY BY AN APPLICATION PROGRAM
-C>   IF THE PROGRAM IS COMPLETELY FINISHED WITH ALL CALLS TO ALL OTHER
-C>   BUFR ARCHIVE LIBRARY ROUTINES, BECAUSE THE MEMORY FREED HEREIN WILL
-C>   RENDER THE LIBRARY AS EFFECTIVELY UNUSABLE FOR THE REMAINDER OF THE
-C>   LIFE OF THE APPLICATION PROGRAM.  HOWEVER, THIS MAY BE A USEFUL
-C>   OPTION FOR APPLICATION PROGRAMS WHICH WANT TO MOVE ON TO OTHER
-C>   UNRELATED TASKS WITHOUT CONTINUING TO TIE UP A SIGNIFICANT AMOUNT
-C>   OF DYNAMICALLY-ALLOCATED HEAP MEMORY RELATED TO THIS LIBRARY.
-C>   OTHERWISE, ALL SUCH MEMORY WILL BE FREED AUTOMATICALLY ONCE THE
-C>   APPLICATION PROGRAM TERMINATES.
```
edwardhartnett, line 21: Seems like you lost a lot of the text here. Was that intentional?

src/bort.f
```diff
@@ -1,75 +1,27 @@
...
-C>                               UFBMEX   UFBMMS   UFBMNS   UFBOVR
-C>                               UFBPOS   UFBQCD   UFBQCP   UFBREP
-C>                               UFBRMS   UFBSEQ   UFBSTP   UFBTAB
-C>                               UFBTAM   UFDUMP   UPDS3    UPFTBV
-C>                               UPTDD    USRTPL   WRCMPS   WRDESC
-C>                               WRDLEN   WRDXTB   WRITDX   WRITLC
-C>                               WRITSA   WRITSB   WTSTAT   CODFLG
-C>                               RDMTBF   INITTBF  STRTBFE  GETCFMNG
-C>                               Normally not called by any application
-C>                               programs but it could be.
```
edwardhartnett, line 67: Great that we can now rely on doxygen to generate this list. I would not really believe it was correct, otherwise...

edwardhartnett (approved):

src/ardllocf.F
jbathegit: Yes, this was intentional.  That text is now part of the docblock for exitbufr.f, which calls ardllocf.F internally.   So the information is still there, but just at a higher level in the hierarchy where the user is more likely to see it ;-)

PR #136, opened 2021-04-20
another batch of Doxygen updates

jbathegit:
I know it's a lot of files, but it's all documentation.  The only exception is that I also moved some function prototypes out of bufrlib.h.in into cfe.h, cread.h and mstabs.h for better encapsulation, but no overall changes to functionality.

Even though there's still plenty to do, this is my last planned push for the next 11.4.0 release, and once it's merged I'll get started on the release.


jbathegit:
BTW, if you want to take a look at the latest Doxygenated documentation for the library, see https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/files.html

edwardhartnett (approved):

PR #143, opened 2021-05-21
master table updates to limit character set to CCITTIA5, including in units and meanings fields

### LINKED ISSUE #140: restrict all Master BUFR Table entries to CCITT-IA5 character set, and allow up to 24 characters for Table B units
WMO recently reiterated that all BUFR element names, unit descriptions and code/flag table entry descriptions are restricted to the CCITT-IA5 (i.e. ASCII) character set, because otherwise they cannot be encoded using Class 0 of Table B.  Many entries in the machine-readable versions of NCEP's current Master BUFR Tables currently contain UTF-8 characters such as subscripts, superscripts and other items that are not part of the ASCII character set, so these will need to be changed.

For the record, it's OK for NCEP to continue to use UTF-8 characters in our master MySQL database as well as in any HTML representations of our Master BUFR Tables; however, files whose contents are meant to be machine-readable and encoded into actual BUFR messages need to be changed to be limited to the ASCII character set.  For example, units fields such as "m s⁻¹" will need to be converted to "m s**-1" in such files.  This will also help with readability when the decoded contents of such files are displayed using library subroutines such as dxdump.

On a related note, the library currently only allows Table B units fields to be up to 14 characters, but WMO regulations allow such fields to be up to 24 characters, so corresponding memory arrays in the library will need to be widened to account for this, and especially considering that many existing units values will now increase in length when changing from the UTF-8 to the ASCII character set.

### LINKED ISSUE #142: generate new v36 master BUFR table files, and set new default output table version to 36
WMO releases new versions of their master BUFR table every 6 months, and these in turn need to be reflected in updates to the tables subdirectory of the NCEPLIBS-bufr project so that the software can read any BUFR message encoded using the new version of the table.  The next version of the WMO master BUFR table will be version 36, effective June 15th, 2021, so we'll need corresponding new NCEPLIBS-bufr v36 master BUFR table files for TableB, TableD and CodeFlag.

Correspondingly, there's also a value in the library that is used as the default table version for any output BUFR message generated by the library.  This default is currently set to a value of 29 (for version 29), and this can be easily overridden by any user within an application program; however, in practice, most users never bother to do this, so after a while there are more and more BUFR messages being generated by operational codes which contain an inaccurate value for this version number once those codes begin making use of more and more new descriptors which are only available in later versions of the WMO master table.  This is especially problematic for any BUFR messages that end up being disseminated outside of NCEP, and for that reason we've found it to be good practice to just go ahead and update this default value every so often within the library, which we'll do again now since it's been three and a half years since the default was set to the current value of 29.

jbathegit:
Fixes #140 

Fixes #142

The output baselines for numerous test cases have been updated to reflect the CCITTIA5 changes as well as the change of the default version number for output files from 29 to 36.  This is implemented via a new bufr-11.6.0.tar file which is downloaded from emcrzdm for use in testing all package updates going forward.

The old bufr.tar file is still available on emcrzdm for use in testing downloads and builds of the recent v11.5.0 release.


jbathegit:
Could someone please take a look at this PR, and let me know if you have any comments or concerns?  Most of the changed files are just regenerated master tables to remove non-CCITTIA5 content.

Thanks in advance for any assistance to get this reviewed and merged! ;-)

jack-woollen:
It all looks good to me. It says its waiting for some tests to finish.

jbathegit:
Hi @jack-woollen all of the 14 tests/checks have passed.  I would just need you or someone else to go into the review section and formally approve it before I/we can merge it.

jack-woollen (approved):

PR #147, opened 2021-06-04
move content of src/list_of_files.cmake into src/CMakeLists.txt

### LINKED ISSUE #42: combine list_of_files.cmake into CMakeLists.txt
It's normal to list all the source code in the CMakeLists.txt file.

I see the impulse to simplify this long list, but putting it in a separate file really doesn't help.

For example, it's typical to do something like:

`find . -name CMakeLists.txt | xargs grep code_file.f
`

to find all the instances of code_file.f in the build system. Putting the list of code files in a separate cmake file just hides it.

jbathegit:
Fixes #42


src/CMakeLists.txt
```diff
@@ -37,7 +37,352 @@ else()
...
+  icvidx.c
+  inittbf.c
+  nummtb.c
+  rbytes.c
+  restd.c
+  sorttbf.c
+  srchtbf.c
+  strtbfe.c
+  stseq.c
+  wrdesc.c)
```
edwardhartnett: I really don't think they each need their own line...

jbathegit: You're right, they probably don't.  If you'd prefer I could put a bunch on each line to reduce the overall number of file lines.

aerorahul:
@jbathegit 

I approve of this change.  I have one suggestion.

In the list of `fortran_src` are files `${CMAKE_CURRENT_BINARY_DIR}/bvers.f` and `${CMAKE_CURRENT_BINARY_DIR}/bfrini.f90`.

If I could make one change, that would be this:

replace `set(fortran_src` with `list(APPEND fortran_src`

and remove those 2 binary files from this section and add them as follows:

```
list(APPEND fortran_src
${CMAKE_CURRENT_BINARY_DIR}/bfrini.f90
${CMAKE_CURRENT_BINARY_DIR}/bvers.f)
```

same with `c_hdr` list.

jbathegit:
OK @aerorahul I can do that.  I'll push up another commit in a few minutes...

aerorahul (approved):
`CMakeLists.txt` changes look good.

PR #155, opened 2021-08-18
use dynamic memory allocation for all builds, and remove _DA suffix

jbathegit:
Fixes #77 

This is a prerequisite for task #78, so that we eventually only have to build one version of the library.  As part of this update, a number of *.F routines no longer require any preprocessing and therefore have now been moved to *.f or *.f90 names.

The problem I'm having is that all of the Python tests are failing, but that's a known issue that pre-dates this PR (see #135) and is therefore unrelated to these changes.

@aerorahul could you please help me to figure out why the Python tests are no longer working?

Issue #135, opened 2021-04-20
python tests don't work when code coverage flags are used?


edwardhartnett:
When I run the code coverage flags with python on, then the python tests fail. They should not.

Are the python tests doing something goofy like looking at text output?

jbathegit:
I think this may be a systemic issue that depends on more than just whether or not code coverage flags are used.  I don't normally run with those flags, because I do most of my work on the WCOSS using the Intel compilers, and in fact I don't even normally test with Python enabled.  But I did try testing this library using `-DENABLE_PYTHON=ON` yesterday on the WCOSS, after manually loading in the python/3.6.3 module.  Everything compiled OK, but all of the python tests still failed during the ctest check, even though I'm not using code coverage flags.

I think we may need @aerorahul to take a look at this when he has some time.

aerorahul:
It is likely that when `creating the python API, the code coverage flags need to be included, and are likely not.
I am not sure I will get to this anytime soon.

PR #156, opened 2021-08-18
Add --output-on-failure to ctest

kgerheiser:
Related to #155 to show error message when tests fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,4 +72,4 @@ jobs:
     - name: test
       run: |
         cd $GITHUB_WORKSPACE/bufr/build
-        ctest
+        ctest --output-on-failure
--- a/.github/workflows/main_new.yml
+++ b/.github/workflows/main_new.yml
@@ -40,7 +40,7 @@ jobs:
     - name: test
       run: |
         cd $GITHUB_WORKSPACE/bufr/build
-        ctest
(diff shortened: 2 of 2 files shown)


aerorahul (approved):

jbathegit:
Interesting that the Python checks are now working.  As noted in #135, @edwardhartnett and I had problems getting these to work even before I tried to submit my PR #155.  I'm really puzzled here.

jbathegit:
I guess I could go ahead and merge this PR back into `develop`, then re-baseline my `jba_noDA` branch and re-run the test and see what happens(?)

aerorahul:
@jbathegit 
If the python tests fail at any point, please open an issue immediately.  
We added those tests to ensure they stay working, and not to be ignored in case they fail.

jbathegit:
> @jbathegit
> If the python tests fail at any point, please open an issue immediately.
> We added those tests to ensure they stay working, and not to be ignored in case they fail.

@edwardhartnett did open an issue for this back in April when he first noticed this (see #135).  But it's still pending/unresolved.

edwardhartnett:
I was just looking at all my issues and thinking that maybe zenhub will help... ;-)

Yes, python tests broke when I added code coverage. Not sure why. But that would have little to do with this issue, which is a cmake option...

PR #158, opened 2021-08-20
code updates to satisfy GNU compiler warnings

jbathegit:
This doesn't address a specific issue in NCEPLIBS-bufr, but rather an overarching issue applicable to all of NCEPLIBS; namely, the desire to have a clean compilation on all compilers without any warnings.

IMHO, in this particular case the GNU compiler warnings were a bit overzealous and unnecessary.  For example, it complained that the following line could produce a potential out-of-bounds condition because ncdesc is a long int and adn is a char*6, even though the conversion specifier limits it to never writing more than 2 digits from ncdesc:

` sprintf( adn, "%c%02ld%c%c%c",
                             adn[0], (long) ncdesc, adn[3], adn[4], adn[5] );`

Similarly, it also complained that the following line could produce a potential out-of-bounds condition because we were using the number of characters in the source string as a limiting factor for the copy, rather than the number of characters in the destination string, and even though units is explicitly defined as char\*10 and card as char*80 within the same source code:

`strncpy( &card[40], units, strlen(units) );`

So again, IMHO the compiler should be smart enough to figure out that there's really no potential hazard here, but whatever, and I went ahead and re-coded these in a different way to git rid of the warnings.  While I was at it, I also went ahead and added Doxygen docblocks to the two C subprograms in question.

### DIFF
--- a/src/restd.c
+++ b/src/restd.c
@@ -1,57 +1,44 @@
 /** @file
-    @author ATOR @date 2004-08-18
+ *  @brief Standardize a local Table D descriptor.
 */
-
 
 #include "bufrlib.h"
 
 /**
-C
-C SUBPROGRAM:    RESTD
-C   PRGMMR: ATOR             ORG: NP12       DATE: 2004-08-18
-C
-C ABSTRACT:  GIVEN THE BIT-WISE REPRESENTATION OF A LOCAL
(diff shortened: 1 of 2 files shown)


aerorahul (approved):
I don't know the details of this code, so happy to approve.

edwardhartnett:
Yes, GNU has gotten tougher with warnings. It's not always a bad thing.

PR #162, opened 2021-09-08
Set CMAKE_C_ARCHIVE_FINISH for macOS

### LINKED ISSUE #161: Need to set CMAKE_C_ARCHIVE_FINISH for macOS
This line runs `ranlib -c` on macOS to make common symbols (i.e. module variables) visible to the linker.

 https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/192731bdeccb1337c90112b34734edba98beafd0/CMakeLists.txt#L44

Need to also add `CMAKE_C_ARCHIVE_FINISH` because there is some C code in the library as well.

I ran into this when building MET and had a bunch of undefined symbols.

kgerheiser:
With C files in the library `ranlib -c` was not being called.

Fix #161


aerorahul (approved):
Good catch @kgerheiser

PR #167, opened 2021-09-28
update function prototypes in bufrlib.h

### LINKED ISSUE #164: updates to bufrlib.h header file
1.  In the C function prototypes, change the types of all of the string length arguments (which are needed when passing string arguments from C to Fortran) from "int" to "size_t".  Corresponding C function calls typically use the return values of strlen() or sizeof() to pass the argument lengths, and those return values are unsigned.  So "size_t" is more appropriate for those prototype declarations since it also unsigned, and newer compilers (e.g. Intel 19 on WCOSS2) will throw errors when trying to map such values into "int" types, which are signed values.
2.  A number of C application programs contain their own UNDERSCORE and prototype declarations for calling library Fortran functions.  Instead, those declarations should be copied into the library header file for global visibility, and applications can then do their own "#include "bufrlib.h" after linking in the appropriate $BUFR_INCx directory.

jbathegit:
Fixes #164

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -19,6 +19,12 @@ guidance from WMO, all fields within all BUFR master table files are now
 restricted to the CCITTIA5 character set.
 [[Issue #140](https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/140)]
 
+* Several prototypes were added to the bufrlib.h header file to enable
+further interoperability with C application programs.  In addition, several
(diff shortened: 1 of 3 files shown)


aerorahul:
@jbathegit Please add @rmclaren as he might be impacted by this work.

rmclaren (approved):

rmclaren:
@aerorahul Fortunately the changes don't affect me.

PR #113, opened 2021-03-05
cleanup some unneeded functionality, and standardize some documentation

jbathegit:
Fixes #107

Issue #111, opened 2021-03-03
consider adding xbfmg utility to NCEPLIBS-bufr utils


jbathegit:
`xbfmg` is a WCOSS utility that breaks apart a file of BUFR messages into files of individual BUFR messages, so that they can be worked with individually rather than as one giant file.

re: [hpc-stack#176](https://github.com/NOAA-EMC/hpc-stack/issues/176), this utility may be useful enough to a wider audience that it makes sense to add it to the NCEPLIBS-bufr utils package, as was done previously with the `debufr` utility.

PR #168, opened 2021-09-28
using Fortran intrinsic functions for character string manipulation, rather than local functions

### LINKED ISSUE #165: change rjust.f from a function to a subroutine
This function right-justifies a string in place.  For some (historical?) reason, it was implemented many years ago as a function rather than a subroutine.  However, at present it always returns a dummy value of zero, which in turn is never used within any of the other library subprograms from which the function is called.  To simplify the code everywhere and stop using unnecessary memory space, it would make more sense to just convert this function to a subprogram which has no return value.

jbathegit:
Fixes #165


jbathegit:
@kgerheiser @rmclaren @aerorahul @edwardhartnett @jack-woollen 

Could one of you please take a few minutes to look at this when you get a chance?  This is a fairly straightforward update, IMHO.

Thanks!

aerorahul:
@jbathegit 
It might seem so, but I don't think we want to manually edit `_bufrlib.pyf`. This is a f2py generated file, although it has been part of the repo since the python API was added, it assumed that the API of the fortran functions didn't change.

@jswhit2 
Do you know if you had to manually edit this file after creating the signature with `f2py -h _bufrlib.pyf`.  Its probably been a while and you were dealing with a frozen version of bufr library, but what would be a better way of keeping the signature and the code consistent with the library changes?

jbathegit:
Understood.  But this isn't a frozen library, so we can't assume that signatures will never change, or that new functions won't ever be added.

FWIW, we've made previous manual updates to `_bufrlib.pyf` when we removed some obsolete code (see #113), and this didn't seem to cause any problems.  But if there's a better way to keep `_bufrlib.pyf` up-to-date, or some automated way we can run f2py when needed to keep things in sync, then I would certainly support that.

rmclaren (commented):
Code changes ok, but why not just use the built in fortran function `adjustr`.

aerorahul:
@jbathegit Yes, that is also what I am advocating.
I know how to generate the signatures with f2py and can easily incorporate it in the build.
I just want to make sure with @jswhit2 that he didn't have to manually edit this signature file (for whatever reason).
Editing this signature could also require updating the python interface in `__init__.py` if there is an API change.

jswhit:
I honestly don't remember if I had to edit the .pyf file originally.  Only way to know for sure is to try generating a new one and see if it works.

jbathegit:
Thanks @rmclaren for the suggestion.  I hadn't seen that particular function before, but it looks promising, so I'll go ahead and give it a try and update this PR accordingly if it works.  Always better to use intrinsics where available and reduce the amount of code you have to maintain within your own library!

jbathegit:
`adjustr` worked as a replacement for `rjust`.  Thanks again @rmclaren for that suggestion!

Similarly, the library also contained it's own `jstchr` subroutine to left-justify a character string, so I've gone ahead and removed that as well and replaced those calls with `adjustl`

I've gone ahead and added these updates to the PR.  Please let me know what you decide re: `_bufrlib.pyf`.  Otherwise, could somebody please review and approve this PR for the time being, and you can always open a separate issue for `_bufrlib.pyf` to automate that into the build at a later date?

rmclaren (changes requested):
Please take a look at the comments I left on https://github.com/NOAA-EMC/NCEPLIBS-bufr/commit/690bec20926d3b64ef49a25d26fd9f85a4c82c2c. Accidentally commented on the commit not the pull request. uug.

jbathegit:
@rmclaren thanks, but for some reason your comments aren't showing up on my screen.  At any rate:

re: 1, yes, this is a weird thing I've noticed with newer versions of vi, that even if you hit the "Tab" button it just moves forward 8 spaces as though you had typed 8 blank characters, rather than actually inserting a tab character in the file.  You can see this if you do a ":set list" in vi.  I've never used any sort of IDE for editing these files; rather, I've always just used vi from the command line, so this is just something with newer versions of vi as opposed to when it actually inserted a tab character.  At any rate, the overall spacing and alignment are still correct inside vi, but these "tab vs. 8 spaces" differences just show up differently in diff displays, which is probably the only reason you noticed this.  That said, I can fix this if you want, but again it's something that only shows up in diff, not in vi.

re: 2, noted and understood, but in this case the second clause has no dependency on the first one, meaning that it will still work regardless of the result of the first clause, so no worries.

aerorahul:
It is showing up in my vi as well. 
Is expandtab on in your vi?

jbathegit:
I think it must be, though it must be set by default because I've never set it myself.  vi is aliased to vim on the wcoss, and I don't have my own personal .vimrc file, which is why I'm assuming expandtab must be set by default inside of vim.

jbathegit:
I just went ahead and fixed the space discrepancies by cutting and pasting tab characters from other lines.  Hopefully this looks better to you now, but it's something that will show up again whenever I edit old code, unless I specifically cut and paste an existing tab character from the same file like I did just now.  So again, not something I want to get in the habit of doing all the time if it only shows up in diff ;-)

jbathegit:
As a reminder, we can always open a separate issue to further investigate the `__init__.py` vs `_bufrlib.pyf` question at a later date, or to automate the building of the latter via f2py.  For now, the manual edits to `_bufrlib.pyf` work fine, and I believe I've addressed all other issues/concerns, so could somebody please approve this PR so I can merge it?  Or else let me know if there's anything else you'd like to see changed?

As a side note, I have a separate PR just about ready to go for #111, but ideally I'd like to get this one resolved first! ;-)

aerorahul (approved):
looks ok.  However, the title of the PR is different than what is actually done in here. (git log will show the PR title)
Suggest editing to:
Using fortran intrinsic functions for character string manipulation, rather than local functions.

jbathegit:
Good point @aerorahul, the gist of this PR has definitely morphed from it's original intent ;-)  I've gone ahead and changed the title as you suggested.

@rmclaren, unfortunately I still can't merge these changes even with @aerorahul's approval, because it still lists you as having requested a change which is blocking the merge.  I assume I've now addressed your concerns, and if so could you please lift this block, or else let me know what else I'm missing?

Thanks!

rmclaren (approved):

rmclaren:
I approved it.

jbathegit:
Got it - thanks everyone!

PR #169, opened 2021-10-01
add xbfmg utility and corresponding tests

jbathegit:
Fixes #111

Part of #32 

I've also included two tests with this new utility.

PR #171, opened 2021-10-15
add a new test_IN program

### LINKED ISSUE #170: subroutine ufbpos only works for BUFR messages containing embedded subset byte counters
Subroutine UFBPOS allows a user to jump forwards or backwards within a BUFR file to a specified message number and subset number within that message.  According to the documentation, this should work on any type of message within a BUFR file; however, during testing, it appears to only work for messages which contain embedded subset byte counters which are encoded using NCEP local descriptors.  In other words, it doesn't work for any messages which don't contain these additional local descriptors, such as compressed BUFR messages.

Based on a brief review of the subroutine logic, it looks like this could be easily fixed by replacing the hardcoded block of logic:

         IBIT = MBYT(LUN)*8
         CALL UPB(NBYT,16,MBAY(1,LUN),IBIT)
         MBYT(LUN) = MBYT(LUN) + NBYT
         NSUB(LUN) = NSUB(LUN) + 1

with a call to subroutine READSB, which includes the same logic block as well as additional logic to handle other types of messages including compressed messages.  But this will need a closer look.

jbathegit:
fixes #170

also includes some minor source code updates to xbfmg utility per discussion in #169

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -53,7 +53,7 @@ writing/encoding BUFR data:
   <th>reading/decoding</th>
   <td>openbf() openbf_f() ufbmem() ufbmex() cobfl()</td>
   <td>readmg() ireadmg() ireadmg_f() readerme() readns() ireadns() rdmemm() readmm() ireadmm() crbmg() datelen() getlens() iupvs01() iupbs01() iupbs3() nmsub() rtrcpt() igetdate()</td>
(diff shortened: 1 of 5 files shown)


jbathegit:
adding this new test_IN program improves the overall code coverage from 66.0% to 68.4% (lines) and 53.1% to 54.6% (branches), as measured by gcovr

jack-woollen (approved):
The changes look okay to me.

PR #173, opened 2021-10-15
remove references to _DA

### LINKED ISSUE #172: bufr-config.cmake is expecting DA targets.
DA targets were removed, but the `cmake/PackageConfig.cmake.in` was not updated.
As a result, `find_package(bufr)` fails with the following error:
```
CMake Error at /opt/modules/gnu-10.3.0/bufr/da5eda3/lib/cmake/bufr/bufr-config.cmake:38 (get_target_property):
  get_target_property() called with non-existent target "bufr::bufr_4_DA".
Call Stack (most recent call first):
  CMakeLists.txt:45 (find_package)


CMake Error at /opt/modules/gnu-10.3.0/bufr/da5eda3/lib/cmake/bufr/bufr-config.cmake:44 (get_target_property):
  get_target_property() called with non-existent target "bufr::bufr_4_DA".
Call Stack (most recent call first):
  CMakeLists.txt:45 (find_package)
```

aerorahul:
Fixes #172

### DIFF
--- a/cmake/PackageConfig.cmake.in
+++ b/cmake/PackageConfig.cmake.in
@@ -1,8 +1,5 @@
 @PACKAGE_INIT@
 
-#  * @PROJECT_NAME@::@PROJECT_NAME@_4_DA - real32 library target with dynamic allocation
-#  * @PROJECT_NAME@::@PROJECT_NAME@_8_DA - real64 library target with dynamic allocation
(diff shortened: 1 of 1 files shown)


jbathegit:
Good catch @aerorahul, and sorry I missed that earlier.  Anyway, looks good!

jbathegit (approved):

PR #175, opened 2021-10-20
increase MXFNLEN to 200 characters in cobfl.c

### LINKED ISSUE #174: increase maximum [path/]filename length for input to COBFL
A WCOSS2 user has asked if we could increase the maximum allowable [path/]filename length of a BUFR file that can be passed in to subprogram COBFL.  The limit is currently set to 100 characters, and the user has requested an increase to 150 characters.

jbathegit:
fixes #174


aerorahul (approved):
Looks straightforward

PR #176, opened 2021-10-28
more documentation updates

jbathegit:
No source or script changes, just more Doxygen markdown ;-)

### DIFF
--- a/src/igetntbi.f
+++ b/src/igetntbi.f
@@ -1,27 +1,26 @@
 C> @file
-C> @author ATOR @date 2009-03-23
-	
-C> THIS FUNCTION RETURNS THE NEXT AVAILABLE INDEX FOR
-C>   STORING AN ENTRY WITHIN INTERNAL BUFR TABLE CTB.
+C> @brief Get the next index for storing an entry within an internal
+C> DX BUFR table
+
+C> This function returns the next available index for storing an
+C> entry within a specified internal DX BUFR table.
 C>
-C> PROGRAM HISTORY LOG:
-C> 2009-03-23  J. ATOR    -- ORIGINAL AUTHOR
-C> 2014-12-10  J. ATOR    -- USE MODULES INSTEAD OF COMMON BLOCKS
+C> @author J. Ator
+C> @date 2009-03-23
 C>
-C> USAGE:    CALL IGETNTBI ( LUN, CTB )
-C>   INPUT ARGUMENT LIST:
-C>     LUN      - INTEGER: I/O STREAM INDEX INTO INTERNAL MEMORY ARRAYS
-C>     CTB      - CHARACTER*1: INTERNAL BUFR TABLE FROM WHICH TO RETURN
-C>                THE NEXT AVAILABLE INDEX ('A','B', OR 'D')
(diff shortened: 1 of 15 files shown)


src/igetntbi.f
```diff
@@ -1,27 +1,26 @@
...
-C>     CTB      - CHARACTER*1: INTERNAL BUFR TABLE FROM WHICH TO RETURN
-C>                THE NEXT AVAILABLE INDEX ('A','B', OR 'D')
+C> @author J. Ator
+C> @date 2009-03-23
 C>
-C>   OUTPUT ARGUMENT LIST:
-C>     IGETNTBI  - INTEGER: NEXT AVAILABLE INDEX IN TABLE CTB
+C> @param[in]  LUN    - integer: Internal I/O stream index associated
+C>                      with BUFR file
+C> @param[in]  CTB    - character: Type of internal DX BUFR table for
```
edwardhartnett, line 13: In my doxygen I get rid of all the extra spaces and hyphen. Also the type is also provided in the doxygen automatically, So I would have:

```
C> @param[in] LUN Internal I/O stream index associated
C> with BUFR file.
```

src/igetntbi.f
```diff
@@ -1,27 +1,26 @@
...
+C>                      entry within CTB
 C>
-C> REMARKS:
-C>    THIS ROUTINE CALLS:        BORT
-C>    THIS ROUTINE IS CALLED BY: RDUSDX   READS3   STBFDX   STSEQ
-C>                               Not normally called by application
-C>                               programs.
+C> <b>Program history log:</b>
+C> - 2009-03-23  J. Ator    -- Original author
+C> - 2014-12-10  J. Ator    -- Use modules instead of COMMON blocks
```
edwardhartnett, line 23: Thanks to a suggestion from Brian, I am now turning these history logs into tabls, like this:

```
 * ### Program History Log
 * Date | Programmer | Comments
 * -----|------------|--------- 
 * 2002-11-01 | Gilbert | Initial.
 * 2009-01-14 | Vuong | Changed structure name template to gtemplate

```

This looks real nice in the output. YMMV.

edwardhartnett (approved):

src/igetntbi.f
jbathegit: Actually, I've found that the type only shows up automatically in Doxygen if the variable is explicitly declared.  However, a lot of variables in this and other legacy NCEP libraries use implicit Fortran typing, e.g. any variable name starting with the letters I-N is implicitly an integer, rather than being explicitly declared.

src/igetntbi.f
jbathegit: OK, I'll take a look at that.  Though I'm not sure how much work it would be to retroactively go back and change that for 150+ subprograms that I've already documented ;-)

src/igetntbi.f
edwardhartnett: OMG that is so evil I had not considered it.

src/igetntbi.f
edwardhartnett: Work is what we do, and keystrokes are free. ;-)

I just do it every time I edit a file for other reasons...

PR #190, opened 2022-02-17
use "-m pip" instead of setup.py for install

jbathegit:


jbathegit:
This is no longer needed, as #199 has now fixed the underlying issue.

PR #191, opened 2022-02-18
Fix Python test by setting PYTHONPATH

kgerheiser:
I don't think this fix is very portable because it hardcodes Python3.8 and 11.6.0, but the tests work. 

Going to open this as a draft for now because I think there's a better way of doing this. 

PYTHONPATH is set in CMake:

```
 set_tests_properties(${_test}
    PROPERTIES ENVIRONMENT "PYTHONPATH=${_install_dir}:$ENV{PYTHONPATH}")
```

So, maybe that needs to be adjusted.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,8 @@ jobs:
         cd bufr
         mkdir build
         cd build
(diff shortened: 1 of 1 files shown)


jbathegit:
This is no longer needed, as #199 has now fixed this problem.

Thanks all the same @kgerheiser  - I sincerely appreciate your efforts on this!

PR #197, opened 2022-03-22
fix test_wrapper_IN script to catch segfaults and other incomplete test runs

### LINKED ISSUE #196: fix test_wrapper_IN script to catch segfaults
This was discovered by accident, but it turns out the existing test_wrapper_IN.sh script doesn't properly catch cases where the underlying test code segfaults or otherwise fails to run to completion.  Instead, it only does a grep of the test code output diagnostics for the string 'FAILED', and if it doesn't find any then it considers the test to be successful.  So the problem is that, if the test code segfaults and doesn't generate any output whatsoever, then the grep for 'FAILED' won't find any matches and will still mark the test as successful.

The script needs to be made smarter, to not only test the output for any 'FAILED' diagnostics, but to also make sure the test code ran successfully in the first place and actually generated output.

jbathegit:
Fixes #196


aerorahul (approved):
looks good.

Issue #125, opened 2021-03-19
Proposed new API for NCEPLIBS-bufr


edwardhartnett:
From @aerorahul :

> Hi Jeff,
> 
> Ron.McLaren is developing an interface to read bufr data (NCEP, WMO) into the JEDI IODA system as part of the ObsProc Reengineering effort based on some simple configuration in a YAML file. We started by writing C++ interfaces around NCEPLIBS-bufr calls, and were able to make fast progress on reading simple BUFR files (files with nonrepeated data, and data contained in fixed repetition sequences). The next stage of this project will involve expanding this idea to support much more elaborate BUFR files (delayed repeated sequences that might be arbitrarily nested). Eventually we hope to support any BUFR file, regardless of construction. The goal is for the user to be able to construct a mapping with minimal knowledge of BUFR. 
> 
> Unfortunately NCEPLIBS-bufr is not well suited for this task. The root problem is that NCEPLIB-bufr is not well encapsulated. The user facing interface expects an inappropriatly detailed level of understanding of the internal data structure and has an elaborate set of data retrieval calls that must applied correctly under specialized circumstances. Added to this there is an elaborate proceedure required for precomputing the number of repetitions for sequences, and this method may not work for all scenarios (especially data retrieved via the ufbevn call). In inspecting user generated code based on NCEPLIBS-bufr we found that users mostly don't precompute the expected data, but rather estimate these numbers which is rather dangerous. Many users we've talked to seem to be very uncomfortable with their understanding of BUFR and the nuances of NCEPLIBS-bufr. We believe its possible to create a simpler, safer, more modern interface, that will be easier to maintain in the long run.  
> 
> None of this work is expected to impact what is in operations currently and your current work and focus is not for nothing. This is the reason to have this meeting at such an early stage so we can identify the areas that need future proofing, come up with a targeted approach for modernizing the code base and working together.
> 
> Thanks,
> Rahul
>

edwardhartnett:
Transferring this converstaion here from email. @aerorahul are you suggesting a completely new set of code? Or some wrapper code for the existing library?

aerorahul:
@edwardhartnett 
We won't be developing any new API or software in this repository, so a discussion here is moot.
A completely new set of code designed from the ground up is the need as will be demonstrated in the discussions.

edwardhartnett:
You will develop and maintain this new code forever? Or you will develop it and hand it off to us to maintain?

[...]

PR #198, opened 2022-03-23
NCEPLIB-bufr Extension for Ioda-Converters Query Interface.

rmclaren:
This pull requests includes the minimal changes I will need to support the query interface for ioda converters (implemented in C++). Mostly its adds methods to expose some of the internal variables of NCEPLIB-bufr via the C interface. 

The most intrusive change I made was to make some of these variables targets. This was done for efficiency reasons, so that I could pass a pointer to the data arrays for message subsets, instead of first copying the data. Copying this data would result in unacceptable performance, so I needed the ability to point to the memory directly.


edwardhartnett:
We need a GitHub issue to describe these changes.

Also the title for this PR is insufficient - what is "Feature/query cxx" supposed to indicate?

jbathegit:
We'll also need one or more test codes which demonstrate the use of these new wrappers and validate them against sample test files.

rmclaren:
I just wrote a long and meandering explanation as part of following issue:
https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/125

I'm not sure if there is a better way to link that issue to this PR... Please go ahead and link them.

rmclaren:
I extended the test_c_interface.c file to cover the new c interface functions.

jbathegit:
Thanks @rmclaren for the test codes.

I still don't fully understand why you need, as you put it, such "intrusive" access to the internal library arrays, but at the same time I don't necessarily have a problem with it if it's really needed, and so long as you're only reading out information and not also updating or otherwise changing anything in those arrays.  To that end, have you already looked at either of the following existing subroutines: [nemspecs](https://noaa-emc.github.io/NCEPLIBS-bufr/nemspecs_8f.html), [nemdefs](https://noaa-emc.github.io/NCEPLIBS-bufr/nemdefs_8f.html), just in case any of those would meet your needs?

Either way, and as you've probably noticed, we're kind of stuck at the moment from being able to merge **any** new updates into the develop branch, until we get some python CI issues sorted out.

rmclaren:
@jbathegit Sorry Jeff, but the functions you mentioned are not helpful...

jbathegit:
This is the next one in the backlog ;-)  I'm rerunning the CI tests now.

While it's not ideal that anyone calling get_inv_f or get_val_f is going to end up with pointers enabling write access to the internal INV and VAL arrays, I get it that it's a performance drag to otherwise have to copy all of that data, so I'm inclined to OK this once the tests pass, unless @jack-woollen has any additional concerns(?)

jack-woollen:
Seems to me like a reasonable way to directly access those data from outside FORTRAN.

jbathegit (approved):

PR #199, opened 2022-03-29
fix failing python tests

### LINKED ISSUE #186: Python tests failing in CI
All the Python tests fail in the CI.

```
Traceback (most recent call last):
  File "/Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/python/test/test_checkpoint.py", line 2, in <module>
    import ncepbufr
ModuleNotFoundError: No module named 'ncepbufr'
```

Also this a bit further up:

```
You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:
    /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/install/lib/python3.8/site-packages
and your PYTHONPATH environment variable currently contains:
    ''
Here are some of your options for correcting the problem:
* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://setuptools.pypa.io/en/latest/deprecated/easy_install.html#custom-installation-locations

Please make the appropriate changes for your system and try again.
```

Not sure why the change, but it seems that ncepbufr is not in the PYTHONPATH.

jswhit:
https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/186

the python module is not being installed by cmake, this is why the python tests fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,7 +69,7 @@ jobs:
         make -j2
         make install
         cd python
-        python setup.py install
+        sudo python3 setup.py install
 
     - name: test
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,6 +68,8 @@ jobs:
         cmake -DCMAKE_INSTALL_PREFIX=./install -DENABLE_PYTHON=ON ..
         make -j2
         make install
+        cd python
+        python setup.py install
 
     - name: test
       run: |


jbathegit:
THANK YOU @jswhit!!

I was hoping it was something as simple as that, and I really appreciate you helping us get this resolved so we can more forward - thanks again!

Fixes #186

jbathegit (approved):

kgerheiser:
Just wondering @jbathegit @aerorahul @jswhit 

Isn't the installation supposed to be done with this block of code, so that `make install` installs Python?

https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/7dbec995765f268fc484b1c080299c49f0005903/python/CMakeLists.txt#L17-L31

kgerheiser:
Just wondering @jbathegit @aerorahul 

Isn't the installation supposed to be done with this block of code, so that `make install` installs Python?

https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/7dbec995765f268fc484b1c080299c49f0005903/python/CMakeLists.txt#L17-L31

jbathegit:
> Just wondering @jbathegit @aerorahul @jswhit
> 
> Isn't the installation supposed to be done with this block of code, so that `make install` installs Python?
> 
> https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/7dbec995765f268fc484b1c080299c49f0005903/python/CMakeLists.txt#L17-L31

I honestly don't know what is the "best" way to do this.  I'm just happy that @jswhit was able to figure out some way to get it working again so we could resume with development and merging in the backlog of other PRs.

These sorts of install questions are above my level of knowledge with Python, but if anyone can come up with a "better/best" way to do this and which still works both locally and in the GitHub runner, then I'm willing to integrate it into the baseline.

PR #202, opened 2022-04-14
Older compiler compatibility for latest bufr C interface code.

### LINKED ISSUE #201: bufr_interface.f90 won't build on orion
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:413.11:

  allocate(jmpb_f, source=jmpb)
           1
Error: Array specification required in ALLOCATE statement at (1)
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:393.11:

  allocate(tag_f, source=tag)
           1
Error: Array specification required in ALLOCATE statement at (1)
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:371.11:

  allocate(typ_f, source=typ)
           1
Error: Array specification required in ALLOCATE statement at (1)
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:350.11:

  allocate(itp_f, source=itp)
           1
Error: Array specification required in ALLOCATE statement at (1)
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:331.11:

  allocate(link_f, source=link)
           1
Error: Array specification required in ALLOCATE statement at (1)
/work/noaa/da/jwoollen/jwfork/NCEPLIBS-bufr/src/bufr_interface.f90:312.11:

  allocate(isc_f, source=isc)
           1
Error: Array specification required in ALLOCATE statement at (1)
make[2]: *** [src/CMakeFiles/bufr_d_f.dir/bufr_interface.f90.o] Error 1

rmclaren:
Some old compilers did not like some of the conventions from the newer c interface code. So some slight modifications were made to make older compilers happier...

### DIFF
--- a/src/bufr_interface.f90
+++ b/src/bufr_interface.f90
@@ -309,9 +309,9 @@ subroutine get_isc_c(isc_ptr, isc_size) bind(C, name='get_isc_f')
   type(c_ptr), intent(inout) :: isc_ptr
   integer(c_int), intent(out) :: isc_size
 
-  allocate(isc_f(size(isc)))
-  isc_f = isc
-  isc_size = size(isc)
+  allocate(isc_f(ntab))
+  isc_f(1:ntab) = isc(1:ntab)
+  isc_size = size(isc_f)
   isc_ptr = c_loc(isc_f(1))
 end subroutine get_isc_c
 
@@ -329,9 +329,9 @@ subroutine get_link_c(link_ptr, link_size) bind(C, name='get_link_f')
   type(c_ptr), intent(inout) :: link_ptr
   integer(c_int), intent(out) :: link_size
 
-  allocate(link_f(size(link)))
-  link_f = link
-  link_size = size(link)
+  allocate(link_f(ntab))
+  link_f(1:ntab) = link(1:ntab)
+  link_size = size(link_f)
   link_ptr = c_loc(link_f(1))
 end subroutine get_link_c
 
@@ -349,9 +349,9 @@ subroutine get_itp_c(itp_ptr, itp_size) bind(C, name='get_itp_f')
   type(c_ptr), intent(inout) :: itp_ptr
   integer(c_int), intent(out) :: itp_size
 
-  allocate(itp_f(size(itp)))
-  itp_f = itp
-  itp_size = size(itp)
+  allocate(itp_f(ntab))
+  itp_f(1:ntab) = itp(1:ntab)
+  itp_size = size(itp_f)
   itp_ptr = c_loc(itp_f(1))
 end subroutine get_itp_c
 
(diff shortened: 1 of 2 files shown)


jack-woollen:
@jbathegit @rmclaren This change was required for building on both orion and hera even with up to date compilers connected. Thanks to Ron for making these updates. I would approve merging.

src/bufr_interface.f90
```diff
@@ -309,7 +309,8 @@ subroutine get_isc_c(isc_ptr, isc_size) bind(C, name='get_isc_f')
   type(c_ptr), intent(inout) :: isc_ptr
   integer(c_int), intent(out) :: isc_size
 
-  allocate(isc_f, source=isc)
+  allocate(isc_f(size(isc)))
```
jbathegit: For this and all of the similar allocations and array copies below (i.e. for link, itp, typ, tag, and jmpb as well), you may want to consider only allocating and copying what you need, which is just the first ntab entries of the jump/link table, rather than the entire allocated size of the table in Fortran, which is 96000 or possibly even more.  For most reading and writing applications, ntab will be much smaller than that maximum number, and don't underestimate the possible runtime drag from having to copy all of those members from one array to another at runtime.  In fact, I'm running into this sort of thing in my work to consolidate all of the _4, _8, _d builds down to one, and where I therefore often have to copy BUFR messages (or portions thereof) from arrays of one build type to another during the interface step.

So, in summary, what you could do is just allocate an array of size ntab and then only copy those first ntab members from the Fortran side.  The value ntab is already stored directly within the moda_tables module, so you wouldn't even need to load a separate module.

jbathegit:
Thanks @jack-woollen and @rmclaren.  For future reference, you always want to have an open issue for any PR, which I see you've already done.  That's great, but then you also want to link the two via the following syntax within a text window of the PR:

Fixes #201

With that, now the PR is linked to that issue, and what's more the issue will automatically get closed once the PR is merged!  Also, you can request me or one or more other persons to be a reviewer by selecting from "Reviewers" in the upper-right corner of the PR, and that way those folks will get automatically notified by email that there's a PR pending review. Or Jack you could have conceivably reviewed and approved this yourself since it's Ron who submitted the PR.

Anyway, all good, and I'm happy to review and approve this myself, but I do have one comment that you may first want to consider about copying large arrays, and which I've annotated in the source diffs.

src/bufr_interface.f90
rmclaren: Thanks for the feedback @jbathegit! I'm not as worried about that for my code as I only read this table data 1 time for each BUFR file. Still I agree it would be good to minimize the resources that we are using, so I'll give your suggestion a shot.

rmclaren: Ok, I made the changes you suggested and tested everything.

jbathegit (approved):

PR #204, opened 2022-04-19
remove .x extension from debufr and xbfmg utilities

### LINKED ISSUE #203: remove .x suffix from debufr and xbfmg utilities
Currently, the debufr and xbfmg utilities are built as executables debufr.x and xbfmg.x.  Instead, the executables should be built without the .x suffixes (i.e., as debufr and xbfmg).  This would make these utilities simpler to find and call from within the install subdirectory, and it would also make them consistent with all of the other utilities which are already built with any .x suffixes.

jbathegit:
Fixes #203 

@jack-woollen I also cleaned up a couple of things with the sinv utility, because it wasn't running cleanly with the Intel compilers on WCOSS2.  Subroutine mtinfo should never be called prior to the first call to openbf, because when that happens then the default master table directory in bfrini will overwrite whatever was passed in previously via mtinfo.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -185,13 +185,13 @@ foreach(test_src ${test_c_interface_srcs})
   endforeach()
 endforeach()
 
-# Test debufr.x utility
+# Test debufr utility
 set(db_flags_1 "-t ../tables")
 set(db_flags_2 "-t testfiles/data -f bufrtab.031 -c")
 foreach(db_num RANGE 1 2)
   set(db_case debufr_${db_num})
   add_test(NAME test_${db_case} COMMAND ${CMAKE_BINARY_DIR}/bin/test_debufr.sh
-           "${CMAKE_BINARY_DIR}/utils/debufr.x ${db_flags_${db_num}}" "testfiles/data/${db_case}"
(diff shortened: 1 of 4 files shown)


jack-woollen (approved):
Thanks for fixing sinv re mtinfo. Good to know. I didn't see a problem running sinv on wcoss2 though. What was the problem you saw? Just curious.  Approved without further ado.

jbathegit:
> Thanks for fixing sinv re mtinfo. Good to know. I didn't see a problem running sinv on wcoss2 though. What was the problem you saw? Just curious. Approved without further ado.

At least when compiling with Intel, I was getting a ctest failure in the test_sinv test.  Looking at the test output, it was failing in mtinfo b/c it was still looking for the master tables in the default directory defined within bfrini, even if I manually changed the 2nd argument to sinv to tell it look for those tables elsewhere.  After hooking it up to the gdb debugger, that's when I realized it was because mtinfo was being called prior to the first call to openbf, which in turn calls bfrini, and where it was overwriting the master table directory location that had previously been stored using mtinfo.

I honestly can't say for sure why it would have worked in your testing, or in the previous GitHub runner tests where it obviously must have worked as well.  All I can say is that it was failing for me using the Intel suite on wcoss2, and it is documented in mtinfo that it shouldn't ever be called prior to the first call to openbf.  Making that fix to sinv ensures it will now work reliably everywhere!

PR #209, opened 2022-05-06
fix C compiler flags for Debug option

### LINKED ISSUE #208: fix DEBUG flags
Currently, building the library using the `-DCMAKE_BUILD_TYPE=Debug` option generates a result that doesn't work with several interactive debuggers, including gdb.  Upon further investigation, it appears that several compiler flags aren't being properly set which are needed to generate symbol tables, and which in turn would allow a debugger session to interactively step through the code.

jbathegit:
Fixes #208


CMakeLists.txt
```diff
@@ -75,9 +75,13 @@ list(APPEND kinds "4" "8" "d")
 
 # Set common to the package compiler flags
 if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -traceback")
   set(CMAKE_C_FLAGS_DEBUG "-O0")
+  set(CMAKE_C_FLAGS_RELEASE "-O3")
 elseif(CMAKE_C_COMPILER_ID MATCHES "^(GNU)$")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -fbacktrace")
```
aerorahul: `-fbacktrace` is valid for Fortran and not C.

jbathegit: OK, is there something comparable for the gcc compiler?

At least for Intel, `-ftraceback` is valid for both the ifort and icc compilers.

kgerheiser: Interestingly enough, I don't think there is.

jbathegit: Thanks Rahul and Kyle!  I'll push up another commit to remove the `-fbacktrace` for the C compiler, so the only remaining flag there will just be `-g`.

aerorahul (approved):
looks good.

PR #216, opened 2022-06-04
allow encoding and decoding of integers larger than 32 bits

### LINKED ISSUE #195: allow encoding and decoding of values larger than 32 bits
It's possible for the library to be asked to encode and decode values which occupy more than 32 bits.  An example of this is the WMO standard Table D BUFR sequence 3-15-010, where an operator is used to encode the constituent Table B BUFR descriptor 0-22-069 (spectral wave density) with 38 bits.

Currently, the only way to encode and decode such a value is by using the _8 build build of the library, which defaults to 8-bytes for **all** declared integers.  However, since a long-term goal is to consolidate all of the various _4, _8, and _d builds down to one (and, ideally, to the equivalent of the current _d build - see #78), then we need to figure out a way to encode values >32 bits in **any** of the current builds, by drilling down into the Fortran code to figure out what internal variables need to be explicitly declared as 8-byte integers, while continuing to allow everything else in the code to be declared as 4-byte integers.  This will allow for maximum library efficacy and portability in the future.

jack-woollen:
Fixes [#195](https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/195)

Part of [#33](https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/33)

Part of [#78](https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/78)

@jbathegit Many thanks for your collaboration and careful review of this work. And especially for making the tests better and more comprehensive.


jbathegit (approved):
Thanks again Jack!

PR #219, opened 2022-06-24
remove MODV references from MODA files and include explicitly where needed

### LINKED ISSUE #218: streamline use of modv modules
Now that all user-configurable arrays within the library are dynamically allocated by default (re: #77), then we no longer need to include `use modv_*` statements within all of the moda* array definition modules.  Instead, we can simplify the overall library design and reduce the overall number of source and preprocessor code lines by moving these `use modv_*` statements directly into any subprograms where the corresponding values are actually referenced.

jbathegit:
Fixes #218

This results in a net reduction of over 200 lines of source and/or preprocessor code from the library.


jack-woollen (approved):
Lots of changes!

jbathegit:
Thanks Jack.  You're right that it's a lot of changed files, but the changes are basically header changes and are similar in nature across all of the files, so I didn't think it would be too overwhelming or unreasonable to include them all in the same PR.

PR #221, opened 2022-06-27
clean up and simplify python/_bufrlib.pyf

### LINKED ISSUE #220: clean up bufrlib.pyf
There are a lot of declarations in python/_bufrlib.pyf that aren't needed, and which in fact are downright misleading given that most of the referenced arrays are now dynamically allocated, rather than being fixed sizes at compile time.  Furthermore, entries are only needed in this file for subprograms which are actually called by the python codes and utilities, and not for every last subprogram in the library.

In short, this python/_bufrlib.pyf file could use a good clean-up.

jbathegit:
Fixes #220

The original version of this file was auto-generated by the f2py utility running on the entire library.  After cleaning it up, it now only contains the interface details for the 30 subprograms that are actually called from within any of the python codes, which in turn eliminates 3700+ unneeded lines from this file.


jbathegit:
Hey @jack-woollen, when you get a chance would you mind giving this a quick look-see?  It's pretty straightforward and doesn't have any direct impact on the unified build stuff, but it's still (I think ;-) a worthwhile cleanup/housekeeping task for the overall NCEPLIBS-bufr project.

jack-woollen (approved):
Sorry, I didn't see the request. Looks good to get rid of the extra stuff and keep it slim and trim!

jbathegit:
Thanks Jack!

PR #222, opened 2022-07-27
Make Type Information Retrievable Via C Interface

### LINKED ISSUE #225: Need some way to get type information on C Interface.
The C interface doesn't have any methods to determine the "type" information for variables. This means that the c/c++ client code needs to assume that all the numeric types are float values when really it is more appropriate to represent them in some other way (int, uint, .... etc).

rmclaren:
This pull request adds the ability to retrieve the type information associated with a mnemonic via the c interface. This information is used to figure out the appropriate type to use to store information. It is associated with the following issue:

https://app.zenhub.com/workspaces/jedi-5eb436d0a544107acce79780/issues/jcsda-internal/ioda-converters/965

ioda-converters issue #965


rmclaren:
We will need to create a new numbered tag when this is merged as well...

rmclaren:
Fixes #225

edwardhartnett (changes requested):
It's not really clear to me what is going on with this PR.

Are you calling C from Fortran, or fortran from C? Then you should be using the ISO compatibility features of fortran. But I don't see that here.

It would be nice to have some comments in this code explaining what is going on.

Is this code relating to handling C/Fortran strings?

rmclaren:
I added some comments to the code.

The new code extends the **bufr_c_interface_mod** who's purpose it is to expose a c compatible API to NCEPLIB-bufr. The import of the iso binding happens at the module level since every function in the module is totally dependent on it. Please see the following:

```
module bufr_c_interface_mod

  use iso_c_binding

  implicit none
```

This is where the types `c_int`, `c_char`, etc come from and why the construct `bind(C, ....)` works.

The blurb `bind(C, name='get_type_info_f')` causes the c compatible version of this function to be created and exposed with the function name **get_type_info_f**. To make it convenient of a C client to use these function declared in this module we provide *bufr_interface.h* which define the signature for all the exposed functions (the client needs only to import this file).

jbathegit:
@rmclaren I'm mostly OK with this; however, I still spotted a couple of inconsistencies in your documentation:

1. It looks to me like the variable unit_str_len is strictly an input (@param[in]) variable, even though you have it listed in the documentation (in both the .c file and the .h file) as an input/output (@param[inout]) variable.  Could you please clarify and tidy that up?
2. I also don't see any documentation (in either the .c file or the .h file) for the variable bits, which should be listed as an output (@param[out]) variable in both files.  Could you please add that as well?

For the record, I'll also note that you could have saved yourself a bit of coding if you had just written your new get_type_info function to internally call the existing library functions nemdefs and nemspecs.  However, I don't think that's something you necessarily need to change at this point; rather, just more of an FYI.

@edwardhartnett I believe @rmclaren has sufficiently answered your concerns; specifically, he's already using the standard ISO binding between C and Fortran, and he's also added some further comments.  So unless you still have some further concerns, and once he's also able to address the 2 minor points I raised above, could you please remove your "change requested" block on this PR?  Otherwise, I don't think that I or anyone else will be able to approve and merge it.

jbathegit (dismissed):

jbathegit:
I've approved this, but @edwardhartnett you still have a block on this PR.  Could you please remove that block, or otherwise let us know what you'd still like to be changed?

rmclaren:
@jbathegit Did a little last minute change to add the description string as part of the type info as well. This will be useful in "print_queries" to indicate what the datafields mean in plain English.

jbathegit:
Please make sure your documentation is correct when you cut+paste.  For example, in the .f90 file, the new variable should be labeled "@param[in] desc_str_len" instead of "@param[in] unit_str_len.  Also, your documentation for the bits parameter says "c_int: reference of element", when it should really be "c_int: bit width of element" in both the .f90 and the .h files.

This may seem nitpicky, but having correct documentation is important, especially when Doxygen renders it as HTML, which is where most users of the library will see it, and where they won't be looking at the source code alongside of it.

jbathegit:
Also, if you're going to continue to make changes to the code anyway, then I reiterate my previous point that you could just have your new routine internally call the existing library functions nemdefs and nemspecs, which already make all of these same internal values visible to application/client codes, including the description string.

rmclaren:
Ok, will do.

rmclaren:
Now wrapping the nemdefs and nemspecs functions in a similar manner we wrap other functions.

jbathegit:
@rmclaren thanks for making these updates!  I like your overall approach, but I still have a few remaining issues/concerns:

- unit_c and desc_c should be declared and documented as "out" parameters in nemdefs_c, rather than as "inout" parameters.
- desc_c is mislabeled in the docblock for nemdefs_c as "unit str" when it should be "description str".
- In the call sequence for nemspecs_c, you're passing in mnemonic_idx but not using it anywhere.  Please replace the 3rd call parameter "1" in the internal call to nemspecs with "mnemonic_idx", in order to actually use that value inside of the routine.
- When calling nemdefs inside of nemdefs_c, why not just directly pass in c_f_string(mnemonic) as the 2nd call parameter and then delete the preceding "mnemonic_f = c_f_string(mnemonic)" statement?  Ditto for nemspecs within nemspecs_c.  You used this same shorthand approach previously with other interface routines (e.g. see table_b_mnemonic in ufbint_c and ufbrep_c), so why not use it here as well?  That way, you don't even need to declare mnemonic_f as a local allocatable in either routine.
- You still have a "use moda_tababd" statement in both nemdefs_c and nemspecs_c.  Please delete those, because you should no longer need access to that module within these two routines.
- Within both nemdefs_c and nemspecs_c, are you sure you want to do an internal call to bort when you get a non-zero return code from nemdefs and nemspecs?  Instead, it might be better if you just pass iret back to the calling program as a call parameter, and that way you can let each calling program check that value and have the flexibility to decide what to do in case it's non-zero.
- There are still a couple of lingering typos where "ujit" should be "unit".

rmclaren:
unit_c and desc_c are inout parameters. The function expects pre-allocated arrays which are then modifies via the `copy_f_c_str` function.

jbathegit:
> unit_c and desc_c are inout parameters. The function expects pre-allocated arrays which are then modifies via the `copy_f_c_str` function.

Yes, the arrays need to be declared and allocated in the calling program, but I don't agree this makes them inout parameters.  IMO, the arrays would only be considered inout if they were required to be initialized with one or more values by the calling program before calling the subroutines, in addition to possibly containing the same or different values upon output.  However, these arrays are only expected to contain one or more actual values on output from those subroutines, and in which case they are strictly out parameters.

In other words, I don't think the fact that array space merely needs to be declared and allocated by the calling program qualifies them as inout.  By that same reasoning, having to declare a non-array variable such as iret in the calling program (in order to have a place for the subroutine to store that return value) would then also require that iret variable to be labeled as inout, which I doubt either of us thinks makes sense.

At any rate, I don't think this is worth quibbling too much over, and all of your other updates look good, so I'll go ahead and approve this, but we're still going to need @edwardhartnett to remove his merge block before it can be merged.  Ed, can you please remove that block or else let us know what you still believe needs to be fixed/corrected in order to do so?

jbathegit (dismissed):

rmclaren:
Converted to draft for now as there are cases where mnemonics are not found by nemspecs_f.

rmclaren:
nemspecs_f seems to be relient on the data that is present. So for example if you have a binary sequence called P__BACKG and that segeunce happens to be missing in the data then nemspecs_f will fail. I really need the information whether any particular sequence is present or not. Should I go back to manually parsing table B if nemspecs_f fails?

jbathegit:
Sorry, I'm not familiar with the sequence P__BACKG.  But if I'm understanding you correctly, you're saying that if the sequence isn't present in the BUFR message (e.g. if it has a delayed replication count of zero), then nemspecs isn't working for any Table B descriptor occurrence within that sequence?

Assuming I'm understanding you correctly(?), can you share some more details, including a sample BUFR message (or file of messages) as well as the table you're using to read it and what specific descriptor you're trying to get the information for?  The reason I'm asking is because I believe that the underlying logic in nemspecs is the same as what you had previously, at least in terms of how it cycles through the internal moda_tababd arrays, so I'm a bit surprised to hear this isn't working for you.  Was this particular case working under your previous code and just not now under nemspecs?

rmclaren:
Yes the "original" version of the function I wrote worked for the all the cases. I don't think nemspecs_f can work in the same way as it indexes to a specific instance of a mnemonic which means it probably tries to take into account the effect of modifying descriptors. Still not sure why it would need to have actual data for this as I thought that relevant meta data was described in the bufr table section (I could be wrong). I'm leaning to just changing things back to the way I had things before... I can rename the original function I had to _get_tabb_type_info_ to make it clear it doesn't account for modifying descritptors.

jbathegit:
Yes, nemspecs is written to focus on a specific occurrence of a mnemonic, since any occurrence of a repeated mnemonic could still potentially be impacted by a Table C operator (e.g. change data width, scale, and/or reference value) that may not be in scope for other occurrences of that same mnemonic.

If you don't mind, I'd still like to see a specific example of a case where nemspecs isn't working, if for no other reason than to see if it might be something that could be easily fixed within nemspecs.  In other words, maybe there's a bug or some oversight in nemspecs that could be fixed or otherwise tweaked to cover your case.

However, if it turns out we can't resolve your non-working example in that way for whatever reason, then I'll support you in going back to your "original" approach.  But in that case I would ask you to not completely discard your nemspecs_f and nemdefs_f updates, because I think they would still be useful for other potential C or C++ codes that may want to use them down the road.  In other words, even if you end up having to go back to a _get_tabb_type_info_ solution for that particular example, I'd appreciate if you/we could still include the latest versions of nemspecs_f and nemdefs_f in your update package for this PR.  That way, it would end up being part of a broader effort to modernize the Fortran/C interface within the library (see open issues #20 and #79), and I think that nemspecs and nemdefs are two important routines to include within that work.

rmclaren:
@jbathegit  Please see your email for a test program and data. You can copy it to your test directory and add it to your CMakeLists.txt file. Don't forget to update the path string to point at wherever you downloaded the data to.

rmclaren:
I added a function called get_tabb_info to get table b data directly.

jbathegit:
For the record, I would have appreciated it if you could have given me a little time (at least 24 hours?) to look at your sample case and try to come up with a better solution before you plowed ahead on this. 

That said, I appreciate that you did keep nemdefs_f and nemspecs_f as I requested.  And after digging into this some more, I do see now why nemspecs_f doesn't handle the particular POE case you provided, since nemspecs_f is looking at what actually exists in the first unpacked subset (via moda_usrint and moda_tables), instead of everything that's possible to be included in any given subset (via moda_tababd).  So that's why it couldn't find POE, since POE didn't exist (the replication count of its parent sequence was zero) in the first unpacked subset.

So your new get_tabb_info will work; however, again, there's a better solution for the logic in that routine, because there are already internal library routines that will handle everything you're doing in the "do idx" loop between lines 628 and 651.  Specifically, you can replace that entire loop of logic with a call to [nemtab](https://noaa-emc.github.io/NCEPLIBS-bufr/nemtab_8f.html) where you pass in LUN along with the mnemonic you're looking for (NEMO), and then you take the output value IRET (assuming it's not zero, meaning it wasn't found in the moda_tababd arrays) and pass that value in as ITAB in a call to [nemtbb](https://noaa-emc.github.io/NCEPLIBS-bufr/nemtbb_8f.html).  Doing that will give you back everything you want except the description string, which you can just get via a direct call to nemspecs as before.

Doing that will dramatically simplify your get_tabb_info logic, plus you'll no longer need to directly import the moda_tababd module, and you'll no longer need the local variables stat_f, idx, scale_str_f, bits_str_f, or reference_str_f either.  So a lot fewer code lines as well, since it keeps get_tabb_info from duplicating code logic that's already available elsewhere in the library.  Please use this approach instead within your get_tabb_info logic, and let me know if you get stuck or otherwise need help for any reason.

rmclaren:
The c interface has been extended with the function calls that were requested , and the get_tabb_info function has been dropped. Please take a look.

jbathegit:
Thanks Ron.  Overall your updates look fine to me, save for a few minor typos.

I'll try reaching out to Ed again to ask him to remove his block from 3 weeks ago, then hopefully we can get this merged for you.  Then after that I can get a new mini-release prepped and tagged and push it out for you.  It will be version 11.7.1

jbathegit (approved):

edwardhartnett (approved):

PR #224, opened 2022-07-28
Link C tests to libm

### LINKED ISSUE #223: C tests need to link to -lm explicitly
I get this error when compiling Bufr with `-DBUILD_SHARED_LIBS=ON` using Ubuntu 20.04 and GCC 10.3.0 in a Docker container.

```
#14 1386.4   >> 2903    /usr/bin/ld: CMakeFiles/test_c_interface_2_4.x.dir/test_c_interfac
#14 1386.4              e_2.c.o: undefined reference to symbol 'round@@@@GLIBC_2.2.5'
#14 1386.4   >> 2904    /usr/bin/ld: /lib/x86_64-linux-gnu/libm.so.6: error adding symbols
#14 1386.4              : DSO missing from command line
```

`test_c_interfaces` makes use of the `round()` function from the C math library, but there is no corresponding `-lm`.

https://github.com/NOAA-EMC/NCEPLIBS-bufr/blob/6c91c657656dcb89df356134c2756de1ceac4ace/test/test_c_interface_2.c#L48-L59

kgerheiser:
Fix #223

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -151,7 +151,7 @@ foreach(test_src ${test_c_interface_srcs})
   set(test_exe ${test}.x)
   add_executable(${test_exe} ${test_src})
   add_dependencies(${test_exe} bufr_4)
-  target_link_libraries(${test_exe} PRIVATE bufr::bufr_4)
+  target_link_libraries(${test_exe} PRIVATE bufr::bufr_4 m)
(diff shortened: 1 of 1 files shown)


edwardhartnett (approved):
Shouldn't you first check that a math library can be found?

kgerheiser:
> Shouldn't you first check that a math library can be found?

I looked into if CMake needed logic to handle `libm`, but this is the correct way. If you have `math.h` there's a corresponding `-lm`. There's no special `find_package()` call needed.

Issue #231, opened 2022-08-30
NCEPLIBS-bufr ctest will fail with jedi/intel-impi/2020.2 module


nicholasesposito:
When building NCEPLIBS-bufr with the jedi/intel-impi/2020.2 module, the process will get all the way through. However, running ctest will cause the first 7 tests to fail.

```
          1 - test_pyncepbufr_checkpoint (Failed)
          2 - test_pyncepbufr_gps (Failed)
          3 - test_pyncepbufr_prepbufr (Failed)
          4 - test_pyncepbufr_rad (Failed)
          5 - test_pyncepbufr_satwnd (Failed)
          6 - test_pyncepbufr_write (Failed)
          7 - test_pyncepbufr_test (Failed)
```

When we build with gnu (jedi/gnu-openmpi/9.2.0) module, everything builds and all the ctests pass.

The full module load process is below:
```
module purge
export JEDI_OPT=/scratch1/NCEPDEV/jcsda/jedipara/opt/modules
module use $JEDI_OPT/modulefiles/core
#module load jedi/gnu-openmpi/9.2.0   # Use this OR intel-impi below
module load jedi/intel-impi/2020.2

module load pybind11
module load bufr
module unload bufr/noaa-emc-11.5.0
module unload json-schema-validator/2.1.0
module unload json/3.9.1
module use -a /scratch2/NCEPDEV/marineda/Jong.Kim/save/modulefiles/
module load anaconda/3.15.1
```

jbathegit:
Just to clarify, what platform are you seeing this on Nick?  And are you seeing any warnings during the Python build step on this particular platform?  If so, then perhaps it's related to #200 which is still awaiting resolution.

I also wonder if this is somehow related to #135, where we've seen similar behavior on other platforms including the WCOSS, and irrespective of whether or not code coverage flags are being used.

Either way, this will probably require assistance from someone more experienced with Python, so I've added some such folks to the thread with the hopes that they can step in to help.

climbfuji:
Can you run the ctests with --verbose --output-on-failure?

Which machine?

nicholasesposito:
This was performed on Hera.
There were no warnings throughout the ecbuild or make steps.

I ran the ctest with the options above. For test 7, the output was this:
```

[...]

PR #243, opened 2022-12-06
CI cleanup; All CI workflows are required to pass before merge to develop.

### LINKED ISSUE #242: improve CI system
- Create a developer build which does FTP testing, address santizer, doxygen build, warning check, code coverage, and Debug build. Turn these settings off in all other builds. Use develop branch of all required NCEPLIBS libraries and the latest version of jasper.
- Create a linux_versions workflow that checks with different versions of dependencies.
- Create a linux_settings workflow that tests different build options with the most recent versions of dependencies.
- Use ubuntu-latest/macos-latest everywhere.
- Use caches everywhere they can be used, including on code checkouts.

edwardhartnett:
Fixes #242 

Cleanup of CI system. All CI workflows are required to pass before merge to develop.


edwardhartnett:
Code coverage is working on developer workflow.

edwardhartnett:
The new Intel build is failing like this:

```
6: ImportError: /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/install/lib/python3.1/site-packages/_bufrlib.cpython-310-x86_64-linux-gnu.so: undefined symbol: __intel_sse2_strcpy
 6/62 Test  #6: test_pyncepbufr_write ............***Failed    0.09 sec

```
Anyone have any idea what is going on?

jbathegit:
This has been an issue for a long time - see #231.  That in turn references #135, where I mentioned this problem back in August of 2021.  The python tests have never worked correctly under Intel, regardless of whether or not the code coverage flags are on.

edwardhartnett:
OK, I will turn off Python testing with Intel...

edwardhartnett:
@jbathegit if this looks good to you, approve it and get it merged. I had to change the settings to match the names used in this PR, so nothing else can be merged without updating their CI files to these. Let me know if that's a problem...

jbathegit:
Thanks for adding in the Intel test, as it's long been a goal of mine to include that as one of the CI tests, but I could never figure out any way to do that!

A question though - is there any particular reason why the python builds and tests are now turned off for the MacOS?  Under the old main.yml, we did have python builds and tests included on MacOS, as well as on all of the various Linux flavors.

edwardhartnett:
@jbathegit Good catch, I will turn on the python tests with MacOS...

.github/workflows/MacOS.yml
```diff
@@ -0,0 +1,53 @@
...
+      with: 
+        path: bufr
+
+    - name: build-bufr
+      run: |
+        cd bufr
+        mkdir build 
+        cd build
+        cmake ..
+        make -j2 VERBOSE=1
```
jbathegit: lines 41 and 42 are now redundant and should be removed

.github/workflows/MacOS.yml
```diff
@@ -0,0 +1,53 @@
...
+        cmake -DCMAKE_INSTALL_PREFIX=./install -DENABLE_PYTHON=ON ..
+        make -j2 VERBOSE=1
+        make install
+        cd python
+        sudo python3 setup.py install
+
+    - name: test-bufr
+      run: |
+        cd bufr/build
```
jbathegit, line 49: I noticed in some of the yml files this was just `cd bufr/build` and in others it was `cd $GITHUB_WORKSPACE/bufr/build`.  Which one is correct?

Even if they both work, I think consistency among all of them would be helpful ;-)

.github/workflows/MacOS.yml
edwardhartnett: Good catch.

.github/workflows/MacOS.yml
edwardhartnett: Fixed.

jbathegit:
I see where in commit 32d0131 you noted you were "taking python out of develop as it clashes with asan".  What is "asan"?

edwardhartnett:
asan is short for the address santizer. It's turned on in develop with -fsanitize=address in C and Fortran flags.

The asan run is great because it catches memory errors and tells you the exact line of the code where the problem is. Super-helpful! But it does not play well with other tools. For example, using the debugger with code that has been built with address santizer causes a similar error to that seen here, a complaint about asan. 

The solution is to do one asan build in develop. I would love to also do a python build in develop, but we are already doing that in Linux, so we can take it out of develop.

jbathegit:
Ahh, OK thanks for clarifying that.  I did see where you'd opened up #244, and that's a bit of a head-scratcher.  Jack or I will have to take a good look at that at some point, but for now I want to get some other stuff tidied up, including #78 which we're very close to resolving.  In the meantime, I'll go ahead and get this merged in.

jbathegit (approved):

Issue #80, opened 2020-11-17
Figure out how to measure Python test coverage


edwardhartnett:
Well I used to know how to do this, so I've got to figure it out again. ;-)

aerorahul:
@edwardhartnett 
I have some experience with `pytest` and `coverage` in python.
I can write a couple of examples and share, but I am not sure I can **take** this ON.

jbathegit:
We should also figure out a way to measure C test coverage.  The gcovr utility doesn't currently measure the C portion of the library either; rather, it only does the Fortran portion.

edwardhartnett:
@jbathegit the spirit is willing, but the flesh is weak. ;-) However gcovr works equally with C code, it just needs to be pointed at the correct subdirectories...

However, if you can figure this out, we will happily use it on the other projects.

jbathegit:
As of #249, this is now resolved for the C portion of the library.

But we still need to figure out how to measure coverage for the Python portion of the library.

jbathegit:
Maybe look into using the pytest-cov package? 

https://pytest-cov.readthedocs.io/en/latest/
https://pypi.org/project/pytest-cov/

PR #249, opened 2022-12-09
enable gcovr for C code, and disable gcovr for test codes

jbathegit:
Part of #80

### DIFF
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -38,7 +38,7 @@ jobs:
         cd bufr
         mkdir build
         cd build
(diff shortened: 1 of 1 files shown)


jbathegit:
So, based on a hunch from looking at the gfortran flags for CMake, I verified that the same options are valid for gcc and tried passing those in as gcc flags for CMake.  And bingo, we now have gcovr coverage for all of the C code in the library as well!

I also included the fix I made earlier on the `r8wraps` branch to exclude the test codes from gcovr, in order to get gcovr working again on that branch.

edwardhartnett (approved):

PR #250, opened 2022-12-14
consolidate down to one _4 build of the library

### LINKED ISSUE #78: Use F90 interfaces and kinds to handle 4/8/d build issues, and only build one copy of the library
This will come after #77 

Currently we build 3 flavors of the library, 4, 8, and D. Instead, use fortran interfaces and kinds to provide one library that handles all three cases.

Not sure when this should happen, we want to try this out on a simpler project first.

jbathegit:
Fixes #78

At long last, this is done!  @jack-woollen please take one last look-see and see if I/we missed anything - thanks!


jack-woollen (approved):
Looking through the changes gives some idea how much effort went into this update, both design and implementation. I don't doubt there may be a wrinkle or two left in the system but that's exactly what application testing will work out. I'd say it's definitely ready for that.

Issue #245, opened 2022-12-06
update Doxyfile.in to doxygen-1.9.1


edwardhartnett:
Doxyfile.in contains setting for the doxygen build.

It needs to be updated from time to time with the ubuntu-latest doxygen release (currently 1.9.1).

Once updated we can use the WARN_AS_ERROR setting to cause doxygen warnings to fail CI. We will do this as a separate issue (see discussion below).

I will add this to #243

jbathegit:
Hi @edwardhartnett.  I agree that Doxyfile.in should be updated from time to time to account for new settings in the doxygen software, so thanks for this!  I'm actually using v1.9.4 on my home desktop.

In principle, I also agree that doxygen warnings should be treated as CI fails.  However, in reality, doxygen warns for a lot of things that it shouldn't.  For example, it thinks the BUFR DX table notation `<PPWSQ1>` is an unsupported xml/html tag even when it's encapsulated inside of a `<pre>` block, so it then flags it with a warning when it really shouldn't, since anything inside of an HTML `<pre>` block is supposed to be treated as plain, preformatted text.  I've seen other warnings as well in the past which have left me scratching my head about how robust doxygen's parser really is, but I haven't had time yet to really dive in and try to figure out a way to get around all of them, nor have I had time yet to even add basic doxygen markup to every subprogram in the library.  So bottom line is that we're nowhere near ready to try and set WARN_AS_ERROR=YES.

edwardhartnett:
In terms of the doxygen version, you want to use the same one as is used in ubuntu-latest (currently 1.9.1). (You can check the version at any time by adding a doxygen --version command to the workflow, or just googling.)

Mostly the version is pretty flexible, but the one time it matters is when doing the doxygen -u to update the Doxyfile. If you use 1.9.4 for this, it may add new parameters to the Doxyfile which will generate warnings on the 1.9.1 run the CI is using, and generate warnings, which will break the build. For me, hilarious confusion ensued until I realized what was going on. ;-)

I will separate the issue of updating the doxyfile from the issue of turning on WARN_AS_ERROR and other warning flags...

edwardhartnett:
This was done and merged. I will close this issue.

PR #251, opened 2023-01-04
updates to doxygen processing

jbathegit:
Part of #245

In addition to cleaning up some doxygen warnings, this update also saves the doxygen output as a new artifact from the developer build.

### DIFF
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -54,3 +54,9 @@ jobs:
         path: |
           bufr/build/*.html
           bufr/build/*.css
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: docs
+        path: |
+          bufr/build/docs/html
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -7,7 +7,7 @@
 * The library has been consolidated into a single build using 4-byte
 integers.  It can still be linked to application codes which are compiled
 using 8-byte integers; however, such codes must now include a call to new
-subroutine setimb8() with a value of .true. before making any calls to
+subroutine setim8b() with a value of .true. before making any calls to
 any other library routines.  Furthermore, since all library functions
 which return integer values (e.g. iupbs01(), isetprm(), ibfms()) will now
 return 4-byte integer values, application codes which are compiled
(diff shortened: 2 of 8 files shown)


jbathegit:
So it turns out the GitHub runner is now using NumPy v1.24, in which the type np.float is no longer valid (it was deprecated beginning in v1.20 and is now throwing an error in v1.24).  I've now changed all of those to np.float64, per the recommendations in https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations

.github/workflows/developer.yml
```diff
@@ -54,3 +54,9 @@ jobs:
         path: |
           bufr/build/*.html
           bufr/build/*.css
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: docs
+        path: |
+          bufr/build/docs/html
```
edwardhartnett, line 62: Neat! I will add this elsewhere...

jbathegit: Thanks.  The benefit here is that we'll always have the updated documentation generated for us automatically, and always using the latest version of Doxygen that's available in the GitHub runner, so that going forward all libraries will use the same version and hopefully be better able to stay in sync with corresponding Doxyfile.in updates.

jbathegit:
@edwardhartnett please let me know if this looks good to you and I'll get it merged.  This cuts the number of Doxygen warnings from 94 down to 9.  The remaining ones will take a little more head-scratching to figure out a workaround.

If you want to see what the corresponding HTML documentation looks like, please see https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_new/index.html  Note that the GitHub runner currently uses Doxygen v1.9.1, as you'd already noted in #245

.github/workflows/developer.yml
edwardhartnett: Yes, really handy! Also note that doxygen is now adding information to the top of the Doxyfile showing what version of doxygen was last used to update it, which is helpful. So right now it says:

`# Doxyfile 1.9.1
`
on the first line of the file.

In my experience it is not necessary to update for every doxygen release. It only needs to be updated if there is a newer feature we want to use (as with WARN_AS_ERROR), or if it starts generating warnings about an out of date Doxyfile.

But it is harmless to update any time you feel like it as well. So feel free to run doxygen -u Doxyfile.in with new versions of doxygen as ubuntu-latest is changed.

Also in many of my developer CI builds I have the command doxygen --version before the build happens. That way if there is a problem I can instantly check what version of doxygen was used by the CI system.

edwardhartnett (approved):

edwardhartnett:
Yes, these changes look great. I think all the extra material in the .md files is really good too.

This is a well-documented library! Nice to see.

edwardhartnett:
BTW while it's terrific and necessary that you fix those warnings, in #253 I turn on another doxygen setting WARN_IF_MISSING which should have been on. It will add many warnings, all of which we must also fix. ;-)

But think how great the docs will look when we finish! ;-)

Issue #246, opened 2022-12-06
Add documentation to all undocumented files, finish conversion of legacy documentation to doxygen (allowing us to turn on WARN_AS_ERROR in Doxyfile)


edwardhartnett:
I'm getting this documentation problem. I remember I've seen it before in other Fortran doxygen projects - not sure if I ever found a solution...

`Error in file /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/utils/sinv.f90 line: 115, state: 22(String)`

Many files in NCEPLIBS-bufr are undocumented. This is a violation of best practices and NOAA coding standards.

edwardhartnett:
It turns out that there is some documentation work that needs to be done in NCEPLIBS-bufr. See discussion in #245.

Documentation has been brought up to a pretty high standard in all the other (active) NCEPLIBS libraries. NCEPLIBS-bufr is one of the most important of the NCEPLIBS libraries, so we want the documentation to look and function at its best. 

The high-level documentation for NCEPLIBS-bufr looks good, and many of the modules and functions look very nice. But some of the code files have only been very roughly converted from the legacy all-caps documentation.

Unlike UFS_UTILS and some of the GRIB repos, NCEPLIBS-bufr never got the surge of documentation help from other team members. Let's correct that, so that NCEPLIBS-bufr can start the new year without a backlog of doxygen work, and with very nice-looking documentation for its next release.

@jbathegit I propose that some of the NCEPLIBS team step in to help get documentation up to a high-standard, as we have done on other NCEPLIBS repos.

The plan:
1. identify and develop one or more examples of good doxygen within the repo. This may take a few PRs to iterate.
2. List the files to be modified in this issue, and divide the work between @AlexanderRichert-NOAA, @Hang-Lei-NOAA , and myself. 
3. We will submit a series of PRs which will be documentation changes only - no code changes. Jeff, your role will be to review and fill in missing information, which we will indicate with a triple-question-mark "???".  When you see that, you need to tell us what should be there instead. 

Doxygen work, once understood, is straightforward and easy. If we each spend just a few hours a week on it, it will be done in no time. This work is very important, but not urgent, we can take our time. WCOSS2 or other operational issues should always take precedence.

What does everyone think?

jbathegit:
Thanks, and this sounds great to me!

[...]

PR #253, opened 2023-01-06
doxygen changes

edwardhartnett:
In this PR I turn on warnings for missing documenation elements in the Doxyfile.in.

I also modify two files to make them similar to other NCEPLIBS documentation. Ask questions about any doxygen change you don't understand.

Part of #246

Note that ALL module variables must be documented. I have added ??? for the missing documentation. @jbathegit what should be there instead? You can answer in code review, or, even better, there is a fancy way that GitHub allows you to change the code itself and suggest that in review as an improvement, which can be accepted with a click.

### DIFF
--- a/src/conwin.f
+++ b/src/conwin.f
@@ -42,7 +42,7 @@ C> Function conwin() works with function invcon() to identify subset
 C> buffer segments which conform to the set of conditions.
 C>	
 C> @param[in] LUN integer I/O stream index into internal memory arrays.
-C> @param[inout] INC1 integer subset buffer start index 
+C> @param[out] INC1 integer subset buffer start index 
 C> @param[inout] INC2 integer subset buffer ending index
 C>
 C> @author Woollen @date 1994-01-06      
--- a/src/conwin.f
+++ b/src/conwin.f
@@ -15,10 +15,10 @@ C> 2014-12-10 | J. Ator    | use modules instead of common blocks.
 C>
 C> @author Woollen @date 1994-01-06
       
-C> This subroutine searches consecutive subset buffer segments
-C> for an element identified in the user string as a conditional node
-C> (i.e. an element which must meet a condition in order to be read
-C> from or written to a data subset). If a conditional element is
+C> This subroutine searches consecutive subset buffer segments for an
+C> element identified in the user string as a conditional node. A conditional
+C> node is an element which must meet a condition in order to be read
+C> from or written to a data subset. If a conditional element is
 C> found and it conforms to the condition, then the internal subset
 C> buffer indices of the "window" (see below remarks) are returned to
 C> the caller for processing.
--- a/src/conwin.f
+++ b/src/conwin.f
@@ -41,9 +41,9 @@ C>
 C> Function conwin() works with function invcon() to identify subset
 C> buffer segments which conform to the set of conditions.
 C>	
-C> @param[in] LUN I/O stream index into internal memory arrays.
-C> @param[inout] INC1 subset buffer start index 
-C> @param[inout] INC2 subset buffer ending index
+C> @param[in] LUN integer I/O stream index into internal memory arrays.
+C> @param[inout] INC1 integer subset buffer start index 
+C> @param[inout] INC2 integer subset buffer ending index
 C>
 C> @author Woollen @date 1994-01-06      
       SUBROUTINE CONWIN(LUN,INC1,INC2)
--- a/src/conwin.f
+++ b/src/conwin.f
@@ -1,66 +1,51 @@
 C> @file
-C> @author WOOLLEN @date 1994-01-06
+C> @brief Search consecutive subset buffer segments
+C> for an element identified in the user string as a conditional node.
+C>
+C> ### Program History Log
+C> Date | Programmer | Comments
+C> -----|------------|---------
+C> 1994-01-06 | J. Woollen | Original author.
+C> 1998-07-08 | J. Woollen | Improved machine portability.
+C> 1999-11-18 | J. Woollen | Increase number of open BUFR FILES.
+C> 2003-11-04 | S. Bender  | Added remarks/bufrlib routine interdependencies.
+C> 2003-11-04 | D. Keyser  | MAXJL increased from 15000 to 16000; unified/portable for WRF.
+C> 2010-04-27 | J. Woollen | correct logical flaw and add documentation.
+C> 2014-12-10 | J. Ator    | use modules instead of common blocks.
+C>
+C> @author Woollen @date 1994-01-06
       
-C> THIS SUBROUTINE SEARCHES CONSECUTIVE SUBSET BUFFER SEGMENTS
-C>   FOR AN ELEMENT IDENTIFIED IN THE USER STRING AS A CONDITIONAL NODE
-C>   (I.E. AN ELEMENT WHICH MUST MEET A CONDITION IN ORDER TO BE READ
-C>   FROM OR WRITTEN TO A DATA SUBSET).  IF A CONDITIONAL ELEMENT IS
-C>   FOUND AND IT CONFORMS TO THE CONDITION, THEN THE INTERNAL SUBSET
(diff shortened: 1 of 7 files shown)


edwardhartnett:
I don't know why this is failing, but I'm pretty sure I didn't break anything! ;-)

jbathegit:
These are failing for the same reasons I noted when I first submitted PR #251  The GitHub runner is now using NumPy v1.24, in which the type np.float is no longer valid.  That type was deprecated beginning in v1.20 and is now throwing an error in v1.24, so I included a patch for that as part of #251.

Since you've now approved that PR, let me get it merged into `develop`, then you can merge `develop` back into your `ejh_docs_1` branch, and then those fails will go away.

src/moda_bitbuf.F
```diff
@@ -9,20 +10,16 @@
```
AlexanderRichert-NOAA, line 10: Just for consistency, maybe author+date should go in the same line?

AlexanderRichert-NOAA (dismissed):

src/moda_bitbuf.F
edwardhartnett: I noticed that as well, and tested it. Turns out that doxygen does the same thing whether those two tags are on the same line as not.

edwardhartnett:
Lets leave this up for a few days into next week so that everyone has a time to check it out and comment or ask questions...

src/moda_bitbuf.F
jbathegit: Did you mean to leave in the original author and date?  It looks like you have that information repeated now in the file, on line 3 (where you've now combined them onto one line) and on lines 8 and 9 (where they originally were, and still on separate lines).  How do you know which one is taking effect in the HTML output?

jbathegit: Never mind, going back now and looking at the [HTML output from the `develop` branch](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_new/moda__bitbuf_8F.html), it looks like the author and date weren't showing up there at all, but I see they are now showing up in the [HTML output from your `ejh_docs_1` branch](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_PR253/moda__bitbuf_8F.html).  So obviously this must have triggered from your new line 3.

So all good, but can you please remove the duplication (i.e. my original version of that information) from lines 8 and 9?

src/moda_bitbuf.F
```diff
@@ -9,20 +10,16 @@
...
 C>        Maximum length of an output BUFR message.
-C>
-C>        @var mbyt
-C>        Length (in bytes) of current BUFR message for each internal
-C>        I/O stream.
-C>
-C>        @var mbay
-C>        Current BUFR message for each internal I/O stream.
-
 	  INTEGER :: MAXBYT
```
jbathegit, line 14: Interesting that it seems the \@var tag isn't really needed here.  Per https://doxygen.nl/manual/commands.html#cmdvar, I had thought that tag was needed to get the variable display shown in the [original HTML output](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_new/namespacemoda__bitbuf.html#a53837bd62186cdc39c8062a88e6489d2), but it looks like you get similar results in the [new HTML output](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_PR253/namespacemoda__bitbuf.html#aeb00e574adb63bf7cc809d540d5a02bc) even if it's not there.    Good to know!

src/conwin.f
```diff
@@ -1,66 +1,51 @@
...
+C> @author Woollen @date 1994-01-06
+      
+C> This subroutine searches consecutive subset buffer segments
+C> for an element identified in the user string as a conditional node
+C> (i.e. an element which must meet a condition in order to be read
+C> from or written to a data subset). If a conditional element is
+C> found and it conforms to the condition, then the internal subset
+C> buffer indices of the "window" (see below remarks) are returned to
+C> the caller for processing.
+C>
```
jbathegit, line 25: Thanks @edwardhartnett for getting the ball rolling for us on this - much appreciated!

As an FYI for you and for all - whenever Doxygen renders the discussion portion of the docblock, it always cuts off after what it thinks is the first sentence and then renders that as its own separate paragraph.  See [here](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html_PR253/conwin_8f.html) and scroll down for an example of how it rendered the first sentence of the above paragraph.

So to make the HTML render a little nicer for this routine, how about we re-write it as _"This subroutine searches consecutive subset buffer segments for an element identified in the user string as a conditional node.  A conditional node is an element which must meet a condition in order to be read from or written to a data subset.  If a conditional element..."_

src/conwin.f
```diff
@@ -1,66 +1,51 @@
...
-C>    THIS ROUTINE CALLS:        GETWIN   INVCON
-C>    THIS ROUTINE IS CALLED BY: UFBEVN   UFBIN3   UFBRW
-C>                               Normally not called by any application
-C>                               programs.
+C> Function conwin() works with function invcon() to identify subset
+C> buffer segments which conform to the set of conditions.
+C>	
+C> @param[in] LUN integer I/O stream index into internal memory arrays.
+C> @param[inout] INC1 integer subset buffer start index 
+C> @param[inout] INC2 integer subset buffer ending index
```
jbathegit, line 46: I may be wrong, but I believe INC1 is actually only ever an output argument and is never an input argument to this routine.  I realize this is how it was listed in the original documentation, but if I'm right it would be nice to correct this as we're going through and updating all of these docblocks anyway.  @jack-woollen knows this subroutine better than I do and can probably tell us for sure.

Either way, I believe it's correct that INC2 is both an input and output argument, but I'll also let Jack confirm that too ;-)

edwardhartnett:
@jbathegit there are some "???" in the module file that I need the documentation for...

edwardhartnett:
Failing to build on MacOS with this error:

f951: Fatal Error: Cannot rename module file 'share_errstr.mod0' to 'share_errstr.mod': No such file or directory
compilation terminated.

What's going on? MacOS why do you hate me?

src/conwin.f
jack-woollen: @jbathegit, true, INC1 is only an outgoing argument of both conwin and getwin.

jbathegit:
> Failing to build on MacOS with this error:
> 
> f951: Fatal Error: Cannot rename module file 'share_errstr.mod0' to 'share_errstr.mod': No such file or directory compilation terminated.
> 
> What's going on? MacOS why do you hate me?

I've seen this before on MacOS, and it makes no sense what this even means or why it's happening, so I usually just click in the output to re-run that one job and it works fine the second time.

src/moda_bitbuf.F
jbathegit: In case you missed my earlier comment, and if we're going to now list the author and date on line 3 in the file block, can we please remove the duplicate author and date tags from lines 8 and 9?

AlexanderRichert-NOAA: I think the idea is to have separate author/date for the file and for the module. I don't love the redundancy but I think this is correct.

jbathegit:
> @jbathegit there are some "???" in the module file that I need the documentation for...

IBAY = Current data subset
IBIT = Bit pointer within IBAY

src/moda_bitbuf.F
jbathegit: Even when the module is the only thing in the file?  Or is this related to the changes being proposed in #252?

AlexanderRichert-NOAA: That's my understanding, yes. If we start putting in multiple subroutines etc. per file (#252), then I think this will make a little more sense :)

edwardhartnett:
OK, I think these changes are ready to merge...

jbathegit (approved):

Issue #262, opened 2023-01-15
program utils/binv.f90 and other programs in the utils directory have zero documentation


edwardhartnett:
In other NCEPLIBS projects we had a lot of legacy documentation to convert, but everything was documented.

In NCEPLIBS-bufr we have programs, subroutines, and modules that don't have a single line of documentation to bless themselves with.

@jack-woollen and @jbathegit the program utils/binv.f90 was written and is maintained by you guys. I can see from the history that it was added two years ago by Jeff, and updated 10 months ago by Jack. At both those times, there was a documentation standard in place. Two years ago it was the legacy standard - 10 months ago it was the improved doxygen standard. Both require that all programmers document all code.

Yet no documentation was ever added. Are documentation requirements not clear?

How come no documentation was even attempted? What is going on with utils/binv.f90?

jack-woollen:
@edwardhartnett @jbathegit I'll definitely get on this, and help out with the other stuff,  in the next couple days. This last couple weeks have been extremely busy with various other projects coming to a head.

edwardhartnett:
OK, there has been a lot of progress. But utils/readbp.f90 is still undocumented.

edwardhartnett:
Looks like all these have been fixed. I will close this issue.

PR #263, opened 2023-01-15
more doxygen: fixing doxygen warnings in all .c and .h files

edwardhartnett:
Part of #262 
Part of #246 

Some more documentation fixes.

With the PR, I think there will be no more warnings for the .c and .h files in the build.

There are ??? in this PR, Jeff and Jack please fill in the missing information.

### DIFF
--- a/src/cfe.h
+++ b/src/cfe.h
@@ -5,7 +5,7 @@
  *  @author J. Ator
  *  @date 2017-11-16
  */
-#define MAX_MEANING_LEN 150
+#define MAX_MEANING_LEN 150 /**< ??? */
 
 #ifdef UNDERSCORE
 #define cmpstia1   cmpstia1_
--- a/src/mstabs.h
+++ b/src/mstabs.h
@@ -2,7 +2,7 @@
  *  @brief Define signatures and declare variables for internal storage
  *  of master Table B and Table D entries.
  *
- *  <p>Since the arrays in Fortran module MODA_MSTABS are dynamically
+ *  Since the arrays in Fortran module MODA_MSTABS are dynamically
  *  allocated and their size isn't known at compile time, then we can't
  *  directly access them from within C.  Instead, we need to allocate
  *  separate array space in C and then use subroutine cpmstabs() to copy
@@ -10,8 +10,7 @@
  *  to these C arrays at run time, in order to be able to access this
  *  information from within C.
  *
- *  @author J. Ator
- *  @date 2014-12-04
+ *  @author J. Ator @date 2014-12-04
  */
 
 #ifdef UNDERSCORE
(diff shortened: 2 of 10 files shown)


edwardhartnett:
@jack-woollen andd @jbathegit can we get this PR reviewed next please?

jack-woollen (approved):
I can work on the ??? in fortran routines once this is merged.

jbathegit:
Sorry, I was tied up with other stuff yesterday.

@jack-woollen can you focus on the binv.f90, sinv.f90.in and cmpbqm.f90 utilities?  In fact, I believe you should be listed as the author on all of those rather than me ;-)

I'll take care of filling in the ??? in the remaining files from this PR and let you know if there's any details I need help with.

jack-woollen:
Absolutely. I have a couple more days of meetings and other stuff but I'll
get on that this week.

src/cfe.h
```diff
@@ -5,7 +5,7 @@
  *  @author J. Ator
  *  @date 2017-11-16
  */
-#define MAX_MEANING_LEN 150
+#define MAX_MEANING_LEN 150 /**< ??? */
```
jbathegit, line 9: @edwardhartnett Is the `<` character really needed here and in other similar definition comments?

In other words, does it really need to be `/**<` instead of just `/**` to start the comment?

src/mstabs.h
```diff
@@ -22,36 +21,47 @@ void cpmstabs( f77int *, f77int *, char (*)[4], char (*)[12], char (*)[4],
         char (*)[24], char (*)[8], char (*)[120], f77int *, f77int *,
         char (*)[120], char (*)[8], f77int *, f77int *, f77int * );
 
-#define MSTABS_BASE(var) mstabs_newCarr_ ## var
+/** ??? @return ??? */
+#define MSTABS_BASE(var) mstabs_newCarr_ ## var
```
jbathegit, line 26: `MSTABS_BASE` is a macro for text substitution within this file.  In other words, it's not a function, so it doesn't have a return value, and neither do any of the below instances where this macro is used.

utils/binv.f90
```diff
@@ -1,83 +1,98 @@
...
+!> @author J Ator @date 2021
+
+!> ???
+!>
+!> @return 0 for success, error code otherwise.
+!>
+!> @author J Ator @date 2021
+PROGRAM BINV
+
+  PARAMETER (MAXSUB=100)
```
jbathegit, line 13: Author should be J. Woollen for this, and for sinv and cmpbqm as well.

utils/xbfmg.c
```diff
@@ -26,13 +26,11 @@
...
 
-#define MXFLEN 125 /* Maximum length of bufrfile [path/]name. */
+#define MXFLEN 125 /**< Maximum length of bufrfile [path/]name. */
 
 /**
  * This function prints program usage information to standard output.
  *   
- * @param prgnam - [path/]name of program executable.
- *
  * @author J. Ator @date 2018-03-01
```
jbathegit, line 34: @edwardhartnett Why was the `@param prgnam - [path/]name of program executable.` comment removed here?  Shouldn't the call parameter for the function be defined here?

src/cfe.h
edwardhartnett: If we use the '<' it means the doxygen comment applies to the thing before, rather than after. So these two are equivalent:

```
/** Comment. */
#define MAX_SOMETHING 2

```
and

`#define MAX_SOMETHING 2 /**< Comment. */`

src/mstabs.h
edwardhartnett: Yes I had to insert those to get doxygen to stop complaining. I will circle back around and see how to better get doxygen to document this macro.

jbathegit: Never mind, in my new PR (still in draft), I've gone ahead and removed that macro altogether.  It's now an unnecessary abstraction since we only have one library build anymore.

src/cfe.h
jbathegit: Ahh, OK, got it

utils/binv.f90
edwardhartnett: A great example of why no undocumented code should be committed to the repo.

I will fix and count on @jack-woollen to add documentation to these codes.

utils/xbfmg.c
edwardhartnett: Doxygen is giving me a problem with this function that I am still exploring. Somehow it is being documented twice. I will circle around to this again when more of the warnings have been cleared up.

utils/binv.f90
jack-woollen: @edwardhartnett  Is there a document explaining how documentation is expected to be done?

edwardhartnett: There are several and I will share them with you on google docs.

But the best way, really, is just to cut and paste an existing example and modify.

The guiding principle is that minimal effort is required from document writers. This lowers the bar as low as possible to encourage programmers to contribute documentation.

Doxygen does all the formatting and makes it looks nice. As a programmer you can focus on the meaning and not the formatting.

PR #268, opened 2023-01-17
Update doxygen for moda_bitmaps.F, moda_bufrmg.F, moda_bufrsr.F

AlexanderRichert-NOAA:
This will need details from bufr team.


src/moda_bitmaps.F
```diff
@@ -1,51 +1,52 @@
...
+C> Data values within this module are stored by subprograms
 C> strbtm(), igetrfel(), makestab() and tabsub().
 C>
 C> @author J. Ator
 C> @date 2016-05-27
 
 	MODULE MODA_BITMAPS
 
-C>        @var nbtm
+C>        @var NBTM
```
edwardhartnett: Instead of using the var tag, document each member variable where it is declared. See [moda_bitbuf.F](https://github.com/NOAA-EMC/NCEPLIBS-bufr/pull/253/files#diff-f27d4964efec136c8497fa36f25a1348c236a03788d95706c0c8962a2c1e47ea) for an example.

AlexanderRichert-NOAA: Fixed

jbathegit:
I filled in everything except for one remaining ???, which hopefully @jack-woollen can fill in at some point ;-)

jbathegit (approved):

edwardhartnett:
Once we are done getting warnings done, a simple grep for "???" will show us all the remaining documentation elements that need to be filled in...

jbathegit:
Yep, that's the plan - fill in everything we can now and just mark all of the unknown stuff with ???, so that way it's easy to go back and find all of them later with a grep.

PR #269, opened 2023-01-18
follow up to fill in missing documentation

jbathegit:
This fills in some missing items from #263

This also completely removes the MSTABS_BASE macro from the library.  This macro abstraction is no longer needed, since dynamic memory allocation is now used for the only remaining library build.

@jack-woollen at some point could you please fill in the documentation for arrays knt, iseq, and jseq in moda_tables.F?  I'm not sure what those are (thanks! ;-)

### DIFF
--- a/src/arallocc.c
+++ b/src/arallocc.c
@@ -11,6 +11,7 @@
 */
 
 #include "bufrlib.h"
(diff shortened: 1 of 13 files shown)


edwardhartnett (approved):
Lots of great changes in this PR. I'm happy to see some of this missing documentation get filled in!

Issue #281, opened 2023-01-20
some build instability in the CI system...


edwardhartnett:
We still are having intermittent build failures. For example the MacOS build just failed for me like this:

```

The following tests FAILED:
	  1 - test_pyncepbufr_checkpoint (Not Run)
	  2 - test_pyncepbufr_gps (Not Run)
	  3 - test_pyncepbufr_prepbufr (Not Run)
	  4 - test_pyncepbufr_rad (Not Run)
	  5 - test_pyncepbufr_satwnd (Not Run)
	  6 - test_pyncepbufr_write (Not Run)
	  7 - test_pyncepbufr_test (Not Run)
Error: Process completed with exit code 8.
```

Then I reran the test, and it worked.

These are all python tests. They fail like this:

```
  Start 1: test_pyncepbufr_checkpoint
Failed to change working directory to /Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/test/testfiles : No such file or directory
```

In the build I see this:

```
Downloading bufr test files...
-- Could not download bufr test files, not building tests
```

Maybe this will be fixed by fixing #241.

edwardhartnett:
Just now, on MacOS, I got this failure:
```

cd /Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/test && /usr/local/bin/gfortran-11 -DINTSIZE_8 -DUNDERSCORE -I/Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/src -I/Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/src/include/bufr_4 -g -fbacktrace -w -fallow-argument-mismatch -fallow-invalid-boz -O3 -isysroot /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.6 -fPIE -fdefault-real-8 -fdefault-integer-8 -c /Users/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/test/test_IN_7.F -o CMakeFiles/test_IN_7_8.x.dir/test_IN_7.F.o
f951: Fatal Error: Cannot rename module file 'share_errstr.mod0' to 'share_errstr.mod': No such file or directory
compilation terminated.
```

[...]

PR #282, opened 2023-01-20
lengthened timeout for test file download

edwardhartnett:
Part of #281 

Let's see if lengthening the timeout helps...


edwardhartnett:
I put some sleep commands in of different times, so that the CI runs would not all use ftp at the same time. But that didn't work, which kind of surprises me...

I will ponder and try something else...

edwardhartnett:
This didn't work so I'm trying something else...

Issue #295, opened 2023-01-24
Tests require documentation, but not doxygen


edwardhartnett:
All tests files are required to be documented.

However, that documentation is never doxygen, because we don't include documentation of our test code within the user documentation for NCPELIBS-bufr.

edwardhartnett:
To be clear, there should be absolutely NO BOILERPLATE text in our comments.

To copy the same comment into every test file is not documentation.

Document each test with a description of what that test does. The description should be different for every test.

edwardhartnett:
The problem with the documentation supplied is that it contains no information. Here's an example from test_IN_1.F:

```
C       This tests a number of library routines and functionality
C       involved in reading input BUFR messages, as noted in the
C       comments and print statements below.
```

There is no information in this comment. ;-)

Let's try again. I will submit a PR and we will try and figure out what to say together...

edwardhartnett:
Thanks for adding more comments to the tests.

PR #296, opened 2023-01-24
better comments for tests

edwardhartnett:
Part of #295 

In this PR I take the doxygen out of some test code comments, and add comments to test code where it is missing.

@jack-woollen and @jbathegit there are some "???" to fill in...

### DIFF
--- a/test/test_IN_4.F
+++ b/test/test_IN_4.F
@@ -1,8 +1,11 @@
 C       This is a test for NCEPLIBS-bufr.
 C
-C       This tests ???
+C       This tests a number of library routines and functionality
+C       involved in reading input BUFR messages, as noted in the
+C       comments and print statements below.
 C
 C       Jeff Ator
+
 #ifdef INTSIZE_8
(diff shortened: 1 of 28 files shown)


jbathegit (approved):

test/test_IN_4.F
```diff
@@ -1,3 +1,11 @@
+C       This is a test for NCEPLIBS-bufr.
+C
+C       This tests a number of library routines and functionality
+C       involved in reading input BUFR messages, as noted in the
```
edwardhartnett, line 4: OK, this is not a good comment if it applies to every test.

Put comments that tell what the difference between these programs are, not just that they are tests.

jbathegit: First of all, this new top-level comment does clearly state whether the test code is for reading input BUFR messages (IN) or writing output BUFR messages (OUT).  So it's not the same top-level comment for every test code.

Secondly, and as each new top-level comment clearly states, all of the specific routines and functionality covered by each test code are already documented below in the existing comments and print statements for that test code.  Do you want me to just cut+paste that same content a second time into this new top-level comment?

PR #298, opened 2023-01-24
fixed utils doxygen

### LINKED ISSUE #297: new doxygen warnings about unsupported xml/html tags caused by inadvertent use of HTML
@jack-woollen your recent changes introduced the following new doxygen warnings (only one shown here):

```
Generating docs for file binv.f90...
/home/ed/NCEPLIBS-bufr/utils/binv.f90:6: warning: Unsupported xml/html tag <bufrfile> found
/home/ed/NCEPLIBS-bufr/utils/binv.f90:6: warning: Unsupported xml/html tag <bufrfile> found

```

The problem is with this text in binv.f90:

`!> Usage: binv <bufrfile> will print bufrfile inventory by message type.
`

Doxygen is trying to interpret '<bufrfile>' as a HTML tag, which it is not.

To fix, we have to add the backslash, which escapes the angle-brackets and causes doxygen to ignore them, like this:

`!> Usage: binv \<bufrfile\> will print bufrfile inventory by message type.
`

Now doxygen will do the correct thing: print the angle-brackets, but not try to interpret them as HTML.

I will put up a PR with the fixed, but I wanted to explain this to you guys so you will know what to do if you see it again. Soon we will have WARN_AS_ERROR turned on, which would have prevented you from committing these changes.

edwardhartnett:
Fixes #297 

@jack-woollen and @jbathegit note how the backslash may be used to escape a character in documentation that otherwise would have special meaning...

### DIFF
--- a/utils/binv.f90
+++ b/utils/binv.f90
@@ -3,7 +3,7 @@
 !>
 !> @author J Woollen @date 1994
 
-!> Usage: binv <bufrfile> will print bufrfile inventory by message type.
+!> Usage: binv \<bufrfile\> will print bufrfile inventory by message type.
 !>
 !> @return 0 for success, error message otherwise.
 !>
--- a/utils/cmpbqm.f90
+++ b/utils/cmpbqm.f90
@@ -3,7 +3,7 @@
 !>
 !> @author J Woollen @date 1997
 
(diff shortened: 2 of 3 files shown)


jbathegit:
Thanks, that's good to know.  FWIW, there are other places in the past where I got around this same problem by using the HTML replacements `&gt;` and `&lt;`

jbathegit (approved):

PR #307, opened 2023-01-31
test documentation

edwardhartnett:
Part of #295 

Documentation for a test.

### DIFF
--- a/src/datelen.f
+++ b/src/datelen.f
@@ -1,28 +1,6 @@
 C> @file
 C> @brief Specify format of Section 1 date-time when reading
 C> BUFR messages.
-
-C> This subroutine is used to specify the format of Section 1
-C> date-time values that will be output by future calls to
-C> any of the BUFRLIB [message-reading subroutines](@ref hierarchy).
-C> 
-C> @author J. Woollen
-C> @date 1998-07-08
-C>
-C> @param[in] LEN --  integer: Length of Section 1 date-time
-C>                    values to be output by all future calls
-C>                    to message-reading subroutines
-C>                    -  8 = YYMMDDHH format with 2-digit year
-C>                           (the default)
-C>                    - 10 = YYYYMMDDHH format with 4-digit year
-C>
-C> <p>This subroutine can be called at any time from within the
-C> application program, and the specified value for LEN will remain
-C> in effect for all future calls to any of the BUFRLIB subroutines
-C> which read BUFR messages, unless a subsequent call is made to this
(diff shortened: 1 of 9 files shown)


edwardhartnett:
Can you tell me a little more about what this test does?

For example, there are a bunch of parameters defined, what do they mean?

Also, it seems like this code just prints error messages in case of test failure. It does not stop execution with a non-zero return code, as a test is required to do.

Am I missing something here?

jbathegit:
I wrote most of these test programs years ago, before the library was part of GitHub and had access to an automated CI runner.  Back then, I would run all of the test codes manually via a single shell script, and the print messages were therefore an easy way to view if there were any problems across any one of multiple test programs run by the script.

That said, and once these test codes were moved to GitHub and the automated CI runner, they'll still fail now if there's any FAILURE message printed - this is done via the test_scripts/test_wrapper_IN.sh script, which runs each test code one at a time and then greps the printed output for any FAILURE string (or for a non-zero return from the code, which can also happen).  So rest assured that any failures are still being properly trapped and alerted, which I agree is required of any valid "test".

As I noted earlier, the specific library features that each code was designed to test is documented in the existing print statements and comments, and I had a bunch of different codes so I just numbered them 1, 2, etc. in the filenames to differentiate between them, as well as using and IN or OUT qualifier depending on whether the test was reading or writing BUFR messages.  There are a lot of test codes since the library has so many different features and ways of doing things, but the total code coverage could still be higher, so I'm planning to add even more in the future ;-)

The declared parameters in this particular test code are just for dimensioning arrays for different library calls that will be made later in the code, so you can easily view the documentation of the library subprogram in question to find out what each value is being used for.  But with all due respect, I really don't think we need to add any more documentation to any of these codes, because they should already be easy enough to follow for anyone who has sufficient knowledge of the library that they'd ever be involved in writing test codes in the first place ;-)

edwardhartnett:
I am involved in this library and even after some study I have no idea what those tests do. ;-) it's common for programmers who work on a project for a long time to think everything is obvious. It's not. I will continue to study that code to try and figure it out. More comments would help a lot.

It's not clear why we would support extra cmake code to scan for a text string instead of using the stop command, which is what is commonly used in fortran unit testing, including all the other NCEPLIBS tests. Let's convert these tests to stop with a non-zero error code in case of failure, and remove the extra cmake code to scan for failure.

edwardhartnett:
@jbathegit there are several places where I've inserted a "???" in the comments. Can you help me understand those sections of code?

jbathegit (approved):

Issue #308, opened 2023-01-31
Cause tests to fail with non-zero return code


edwardhartnett:
This can be very easily accomplished by inserting a stop command after every error message.

jbathegit:
Since you've been working on the IN codes, I'm going to start working on the OUT codes so we don't step on each others toes.

jbathegit:
But here's the thing - the OUT tests are writing BUFR output to test the output side of the library.  And the way each of those tests works is that there are few, if any, stops inside of the code, and instead the full test can't be judged a success or failure until it has run through completely and the resulting output is then compared (using `cmp -s` or similar command in a UNIX shell wrapper) to an expected baseline result to confirm it matches the expected output.  So how do we replicate that functionality inside of the code if all we're looking at is the return code to gauge success or failure?

jbathegit:
I've been snooping around for ideas, and it seems like there's now a `call execute_command_line()` intrinsic that can be called from inside of Fortran, at least as of the Fortran 2008 standard.  It's modeled on the old `system()` function in C, and it can return the exit status of a shell command (e.g. from 'cmp -s') as argument exitstat right within the flow of the code, as long as it isn't allowed to run asynchronously.

So it seems like that might do the trick, unless you know of any potential drawbacks or can think of anything I may be missing here(?)

jbathegit:
The `execute_command_line()` intrinisic worked great, so now we have an easy way forward for all of the OUT tests.  I've just completed the transitions for OUT_1 and OUT_2, and I'll continue with the remaining ones tomorrow.

edwardhartnett:
Great to see this accomplished and the test situation simplified.

Issue #244, opened 2022-12-06
most or all tests fail with address santizer turned on....


edwardhartnett:
We should be testing with the address santizer to ensure we do not have memory leaks. But when I turn it on, most or all of the tests fail. (Sample output below).

We need address sanitizer to work, so we need to fix all these memory leaks. Our code should never never never leak memory; in this case it might be OK, because these might just be memory leaks in the tests, not the library code. But we need to fix them so that we will be able to spot memory problems in the code.

To use address santizer, build with GNU and add the option: -fsantize=address.

```
=================================================================
==11427==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55e08134a6a0 at pc 0x55e0812c8b1c bp 0x7ffc1340e820 sp 0x7ffc1340e810
READ of size 8 at 0x55e08134a6a0 thread T0
    #0 0x55e0812c8b1b in iupm_ /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/src/iupm.f:66
    #1 0x55e0812b76cc in wrdlen_ /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/src/wrdlen.F:110
    #2 0x55e0812b4d1c in openbf_ /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/src/openbf.f:200
    #3 0x55e0812a6047 in cmpbqm /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/utils/cmpbqm.f90:48
    #4 0x55e0812a28c6 in main /home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/utils/cmpbqm.f90:119
    #5 0x7fda11fd3d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)
    #6 0x7fda11fd3e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f)
    #7 0x55e0812a5774 in _start (/home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/build/utils/cmpbqm+0x3b774)

0x55e08134a6a1 is located 0 bytes to the right of global variable '*.LC2' defined in '/home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/src/wrdlen.F' (0x55e08134a6a0) of size 1
[...]

PR #313, opened 2023-02-04
convert test_IN_1 to F90, allow it to return non-zero return code on error

edwardhartnett:
Part of #308.
Part of #314


test/CMakeLists.txt
```diff
@@ -3,6 +3,17 @@
 #
 # Rahul, Jeff Ator, Ed Hartnett
```
jbathegit, line 5: Maybe add Rahul's last name here for consistency.

test/CMakeLists.txt
```diff
@@ -118,6 +129,11 @@ list(APPEND test_kinds 4 8 d)
 
 set(fortran_8_defs INTSIZE_8)
 
+# Add tests which don't require output log processing.
+foreach(kind ${kinds})
+  create_test(intest1 ${kind})
+endforeach()
+
 # IN tests
```
jbathegit, line 137: This needs to be `foreach(kind, ${test_kinds})`, so that each test code gets built and tested as _4, _8, and _d (i.e. 3 CI tests per test code, just like for all existing IN and OUT tests).  That way, we confirm that each of these test codes still works correctly under each of the 3 possible existing build scenarios.  It's only the library itself that's now consolidated down to one build, which is why we need to hardcode `bufr_4` above.

test/CMakeLists.txt
```diff
@@ -3,6 +3,17 @@
 #
 # Rahul, Jeff Ator, Ed Hartnett
 
+# This function builds, links, and runs a test program.
+function(create_test name kind)
+  add_executable(${name}_${kind} ${name}.F90)
+  target_link_libraries(${name}_${kind} PRIVATE bufr_${kind})
+  set_target_properties(${name}_${kind} PROPERTIES COMPILE_FLAGS "${fortran_${kind}_flags}")
```
jbathegit, line 10: Per below, this should be `PRIVATE bufr_4` (or `PUBLIC bufr:bufr_4`).  In other words, no `${kind}` option here in the target_link_libraries action, because there's only one build of the library now.

.github/workflows/MacOS.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         mkdir build 
         cd build
         cmake -DTEST_FILE_DIR=/Users/runner/data -DCMAKE_INSTALL_PREFIX=./install -DENABLE_PYTHON=ON ..
-        make -j2 VERBOSE=1
+        make VERBOSE=1
         make install
         cd python
```
jbathegit, line 50: I did see your #314, but for consistency should we also remove the `-j2` from the other yml scripts as well?  Or were you waiting to see if that resolved the problem on MacOS before making this change across the board?

test/intest1.F90
```diff
@@ -0,0 +1,100 @@
...
+  ! Read and check some long character strings.
+  call readlc(11, smidstg, 'SMID')
+  call readlc(11, softvstg, 'SOFTV')
+  IF (smidstg(7:9) .ne. 'UAO' .or. softvstg(5:12) .ne. '5.8.5.10') stop 13
+
+  ! Close the test file.
+  call ccbfl()
+
+  print *, 'SUCCESS!'
+end program intest1
```
jbathegit, line 105: Just so I understand, and since this is no longer being run by the test_wrapper_IN.sh script, the idea is that if it makes it all the way through to the end, then it will end up returning with a zero return code by default, but if it hits any of the prior `stop` statements then it will end up with the associated positive number as a return code, and in turn that will definitely get caught by the CI runner and the test will fail, correct?

jbathegit: Also, I noticed you started with `stop 2` as your first error exit rather than `stop 1`.  Any particular reason for that?  Not a big deal (assuming I'm correctly understanding that anything other than zero automatically gets trapped as a failure?), but more just curious than anything.

test/CMakeLists.txt
```diff
@@ -3,6 +3,17 @@
...
+  set_target_properties(${name}_${kind} PROPERTIES Fortran_MODULE_DIRECTORY
+    ${CMAKE_CURRENT_BINARY_DIR}/include_${kind})
+  add_test(NAME ${name}_${kind} COMMAND ${name}_${kind})
+  target_compile_definitions(${name}_${kind} PUBLIC -DKIND_${kind})  
+endfunction()
+
 # Fetch test data from: https://ftp.emc.ncep.noaa.gov/static_files/public/bufr.tar
```
jbathegit, line 17: Related to the below discussion about each test code being built and tested as _4, _8, and _d, shouldn't this flag to target_compile_definitions be `PRIVATE "${fortran_${kind}_defs}"` as it is in the IN tests section, rather than `PUBLIC -DKIND_${kind})`?  Or do these end up evaluating to the same thing?

In other words, what does `PUBLIC -DKIND_${kind})` end up doing here?

test/CMakeLists.txt
edwardhartnett: OK

test/CMakeLists.txt
edwardhartnett: OK, let me take a look...

test/CMakeLists.txt
edwardhartnett: OK I will fix.

.github/workflows/MacOS.yml
edwardhartnett: I intend to put the j2 back in, as soon as the rest of the tests are converted.`

test/intest1.F90
edwardhartnett: Correct. If the program reaches "end program" then it will return a 0.

If any stop is called, it will return non-zero. From the number of the stop, you can tell exactly where the failure is, if you use a different number for each stop command.

Why not start with 1? I don't know. Good question. I just started with 2 for some reason.

test/CMakeLists.txt
edwardhartnett: Let me take a look and get back to you. I copied this from another NCEPLIBS...

test/CMakeLists.txt
edwardhartnett: Done.

test/CMakeLists.txt
edwardhartnett: Fixed.

test/CMakeLists.txt
edwardhartnett: OK, looks like the only preprocessor flag set in existing tests is INTSIZE_8 for _8 builds.

There doesn't seem to be anything set for _d builds or _4 builds.

So what this does:
  target_compile_definitions(${name}_${kind} PUBLIC -DKIND_${kind})  

is for each test, defines either KIND_4, KIND_d, or KIND_8.

edwardhartnett:
What is the purpose of INTSIZE_8?

In the code we have:


```
#ifdef INTSIZE_8
        INTEGER*4       IREADSB, IUPBS01, IUPBS3, IBFMS
        INTEGER*4       MXBF, nbyt, ierr
#endif

```

If the code is compiled with INTSIZE_8 then IREADSB is defined as INTEGER*4.

So when compiled without INTSIZE_8 it will be the default int size, again INTEGER*4.

But when I did not do this, my test still works...

test/CMakeLists.txt
jbathegit: Correct, there are no preprocessor flags for _d or _4 builds, because none are needed.

It's fine with me if you want to replace INTSIZE_8 with KIND_8 moving forward.  I presume the latter is just to keep things consistent with other libraries or with general industry practice?  Or just a personal preference?

Also, is the `set_target_properties(${name}_${kind} PROPERTIES Fortran_MODULE_DIRECTORY` line really needed here?  I've never had to use that before.  If it's really needed then fine, but it isn't used anywhere else for any other builds in this file, so if it's not really needed then my preference would be to keep things simpler and just remove it.

jbathegit:
> What is the purpose of INTSIZE_8?
> 
> In the code we have:
> 
> ```
> #ifdef INTSIZE_8
>         INTEGER*4       IREADSB, IUPBS01, IUPBS3, IBFMS
>         INTEGER*4       MXBF, nbyt, ierr
> #endif
> ```
> 
> If the code is compiled with INTSIZE_8 then IREADSB is defined as INTEGER*4.
> 
> So when compiled without INTSIZE_8 it will be the default int size, again INTEGER*4.
> 
> But when I did not do this, my test still works...

As noted above, the INTSIZE_8 is just a preprocessor flag for when we're building with 8-byte integers, similar to the KIND_8 you're now using.  The reason it's used here is because IREADSB, IUPBS01, IUPBS3, and IBFMS are all library functions which return an integer, and since the only library build is now _4, then it's always going to return a 4-byte integer, even when the application program is compiled as _8.  I'm surprised to hear you didn't encounter any errors when you removed those explicit INTEGER*4 declarations, because I was definitely seeing them previously on some platforms when Jack and I were working on #78, and which is why I had to put those declarations in in the first place.  I agree they're not needed when you're testing a _4 or _d build of the test code, but when you're building any application code as _8 then you're going to have a mismatch if the code is expecting an 8-byte integer to be returned and instead it only gets back a 4-byte integer.

Similarly, MXBF, nbyt, and ierr are all arguments to the C library function crbmg.c, and the prototype is explicitly defined with these as "int" arguments, which again equates to 4-byte integers since we're only doing 4-byte builds of the library anymore.

So bottom line, I think those explicit INTEGER*4 declarations need to be retained, and again I'm really surprised this didn't trip you up at all during any of your testing.

edwardhartnett:
OK, I have fixed those types. Seems like they are always going to be integer 4, so we don't need the ifdef, we can just always define them as int*4.

test/CMakeLists.txt
edwardhartnett: OK, actually it seems that no preprocessor definition is required, since the variables can be declared integer 4 in the test and that will always work.

I have removed the fortran module directory line. This is not needed here because we are not using modules.

edwardhartnett:
OK, @jbathegit take another look and see if we can get this merged...

jbathegit:
I'm looking at it right now - just give me a couple of minutes...

jbathegit:
So I'm curious why the `target_compile_definitions(${name}_${kind} PUBLIC -DKIND_${kind})` preprocessor def is no longer needed in test/CMakeLists.txt?  How else would the `#ifdef KIND_8` block get triggered in intest1.F90 when it's testing the _8 build of that code?

edwardhartnett:
Your right, of course. I forgot that. Let me re-add it...

jbathegit:
Thanks, but shouldn't that def come before (vs. after) the `add_test` directive?  Or does it not matter?

Either way, I'm really stumped how the intest1_8 test wasn't failing before you added that line back in.

edwardhartnett:
If we could get the memory errors in the library fixed, then presumably the address sanitizer would find a problem without the KIND_8 flag.

My idea is to convert all the tests, remove the complexity related to testing in the CMakeLists.txt file, and then, with a more simple test situation, return to the problem of the memory issues.

jbathegit:
OK, fair enough.  I can help some with this, and hopefully @jack-woollen can help us out too, especially with the memory issues noted in #244 and #319.  But I also need to get started with #79.

jbathegit (approved):

Issue #254, opened 2023-01-06
Change all code to F90 free format, with an extension of .F90 everywhere


edwardhartnett:
On other NCEPLIBS projects I have converted everything to F90 and it was surprisingly easy and surprisingly helpful.

To do it, just use git mv to change the filename, change the CMakeLists.txt, and then in the file, change the way lines are continued.

The benefit is that free-formatted code is easier to work with and looks better. Also the use of the extension "F90" everywhere also allows for pre-processor use, when and if needed. And it removes the common mistake of referencing a code file with the wrong extension, when multiple extensions are in use.

Another benefit (which is what I first noticed) is that doxygen is able to find the types of subprogram parameters automatically for F90 code, but not for F77 code.

@jack-woollen and @jbathegit please consider...

edwardhartnett:
I believe we all agree that all new files added should be .F90 and in Fortran free format.

We can convert existing files after testing is complete.

edwardhartnett:
OK, we have agreed to do this over time, and only when testing is fully in place for the affected code. This will take a good long while...

edwardhartnett:
@jbathegit what do you think about this issue?

jbathegit:
I agree that this is a worthwhile goal, but just as I noted in #328, I also haven't done any further work on this, because I've been prioritizing my time towards improving the level of testing coverage.

Note that for many files and routines we should be able to address both of these issues simultaneously.  But that said, and for cases where changing to F90 free form may be a bit more time-consuming, I do think that #328 should be a higher overall priority because of its value in diagnosing undeclared variables that result from typos.

edwardhartnett:
The good news is that converting to F90 is pretty easy. IIRC the only things that have to be done are:
1 - change old-style line continuations to new-style
2 - look for and change "DOWHILE" to "DO WHILE".

At that point the file can be reformatted. I have been thinking we should get some autoformator for NCEPLIBS Fortran code, so we have similar and consistent indents etc. However, in the absence of that I've just run the emacs Fortran formatter, which does a pretty good job. Perhaps your editor has a similar capability.

It's not necessary to convert all at once, some files can be .f, some .F90.

When this is done, make sure you use .F90 not .f90, because we will certainly want to use the pre-processor and it's confusing to have some files f90 and some F90.

I agree this is not a priority, just wanted to see what you thought of this issue.

PR #317, opened 2023-02-11
consolidate all moda and modv files into single files

### LINKED ISSUE #233: rename all moda*.F files as moda*.f90
Since all builds of the library now use dynamic allocation by default (as of v11.6.0+), then the moda*.F files no longer contain any pre-processing code since that time, and therefore these same files no longer need to be run through the pre-processor during the build step.  Renaming these files to moda*.f90 will automatically bypass the pre-processor for these files during the build step.

As a side note, the only Fortran library files which still contain any pre-processing code, and which therefore still need to be run through the pre-processor during the build step, are irev.F and wrdlen.F

jbathegit:
Fixes #233

Part of #252

Part of #254


edwardhartnett (approved):
This is very helpful in reducing the number of files!

PR #318, opened 2023-02-14
removed program history logs

### LINKED ISSUE #304: Remove Program History Logs
At a recent discussion we agreed to remove the program history logs, for the following reasons:
* They are bulky, taking up most of our documentation, yet provide no value to users.
* They are repetitive, with many history items appearing in many files.

We will remove them in one operation and one PR, so that if anyone wants them, they can go back to one commit in the repo and get them all.

edwardhartnett:
Fixes #304


jbathegit:
Thanks Ed for all of your work on this, and with all of the documentation updates in general to help us get this library where it needs to be!

jbathegit (dismissed):

jbathegit:
OK, so why am I now getting the build error `/home/runner/work/NCEPLIBS-bufr/NCEPLIBS-bufr/bufr/utils/xbfmg.c:28: error: argument 'prgnam' from the argument list of prtusage has multiple @param documentation sections (warning treated as error, aborting now)` when I didn't even touch xbfmg.c?

Is it because I modified debufr.c.in to put in the correct date in the prtusage docblock for that utility, and now the compiler somehow thinks they're different definitions for a single prtusage routine, even though they're two different routines and separately defined within each source file?  How would the compiler even know about the one in debufr.c.in when it's compiling xbfmg.c?

edwardhartnett:
Let me take a look...

edwardhartnett:
OK, doxygen was getting confused by the prototypes of these functions.

No need for prototypes for functions you are about to define. Prototypes are only needed for functions that you are using in a code file before they are defined.

I removed the prototypes and now everything works.

jbathegit (approved):

PR #322, opened 2023-02-15
Converting test_IN_2 to F90.

edwardhartnett:
Part of #308 
Part of #295 

Converting test_IN_2 to F90.


jbathegit:
I removed the comments which referred to the old test codes, since those no longer exist and might confuse anyone who tries to look for them.

I also added a missing `add_dependencies` line to CMakeLists.txt, b/c we always want to insure that the library gets completely built first before any of the test codes, just like we have for all of the other test codes and utilities.  And I removed the `set_target_properties ... Fortran_MODULE_DIRECTORY` line that we'd agreed to remove earlier, but which somehow made it's way back into the file.

jbathegit (approved):

PR #324, opened 2023-02-16
testing IN_3

edwardhartnett:
This is failing, but only for the _8 version of the test.

Yet the original test reports no failure. 

What's going wrong I wonder?


jbathegit:
Where is the intest3.F90 code?  I don't see it in your branch.

jbathegit:
There's no source code included in your commit, only the CMakeLists.txt update.  That's why all of the tests are failing.

jbathegit (approved):

PR #327, opened 2023-02-18
convert test_IN_4.F to F90

edwardhartnett:
Part of #308
Part of #295 
Part of #254 

Converted test_IN_4.F to F90.


test/intest4.F90
```diff
@@ -0,0 +1,87 @@
...
+  call setim8b(.true.)
+#endif
+
+  ! Open the test file.
+  call cobfl(filnam, filost)
+
+  ! Set date format.
+  call datelen(10)
+
+  ! Open /dev/null for output. ???
```
edwardhartnett: Jeff what is going on with opening /dev/null on unit 11? What is it used for?

jbathegit: Since for this test we're using cobfl to read in a BUFR message directly from filnam, then we won't actually be reading from the logical unit pointed to via openbf.  However, we still need to call openbf to set io = 'SEC3' (to tell it to decode messages according to Section 3), and since we still need to call openbf, then we still need to pass in a logical unit number that points to some file on the system.   So, bottom line, since in this case we'll never actually try to read anything from that logical unit, then /dev/null works as well as anything and avoids having to otherwise create an empty dummy file that we'll never read from.

I realize this can be confusing for anyone who's not familiar with all of the different ways that BUFR messages can be input to the library, so I'll go ahead and document this in the code.

jbathegit (approved):

PR #332, opened 2023-02-25
converting test_IN_5.F to intest5.F90, test_IN_6.F to intest6.F90

edwardhartnett:
Part of #308
Part of #254 

Converting IN tests to F90, using stop to return non-zero error code on failure of test.


edwardhartnett:
I will take this down because these are already part of #331 ;-)

PR #374, opened 2023-03-13
testing bort()

### LINKED ISSUE #377: tests for bort() and bort2()
We need tests for bort() and bort2().

See discussion in #340.

edwardhartnett:
Part of #33 
Fixes #377

Testing bort() and bort2().

This PR also provides the way to test other bort() calls. In the program test_bort.F90, each abort case can be set up and run.

### DIFF
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -6,11 +6,11 @@
 # Ed Hartnett 3/12/23
 
 ./test_bort_4 bort 1
-[ $? != 1 ] &&  return 1
+[ $? != 1 ] &&  exit 1
 
 ./test_bort_4 bort 2
-[ $? != 1 ] && return 1
+[ $? != 1 ] && exit 1
 
 # If we made it here, all error codes were correctly returned, and the
 # test passed!
-return 0
+exit 0
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -1,0 +1,21 @@
+# This is a test script for NCEPLIBS-bufr.
+#
+# This script tests aborts.
+#
+# Ed Hartnett 3/12/23
+
+./test_bort_4 bort 1
+if [[ $? != 1 ]]
+then
+    exit 1
+fi
+
+./test_bort_4 bort 2
+if [[ $? != 1 ]]
+then
+    exit 1
+fi
+
+# If we made it here, all error codes were correctly returned, and the
+# test passed!
+return 0
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -5,10 +5,10 @@
 # Ed Hartnett 3/12/23
 
 ./test_bort_4 bort 1
-[ $? != 1 ] &&  exit 1
+[ $? != 1 ] &&  return 1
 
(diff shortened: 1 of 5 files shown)


test/run_test_bort.sh
```diff
@@ -0,0 +1,16 @@
...
+# This script tests aborts.
+#
+# Ed Hartnett 3/12/23
+
+./test_bort_4 bort 1
+[ $? != 1 ] &&  exit 1
+
+./test_bort_4 bort 2
+[ $? != 1 ] && exit 1
+
```
jbathegit, line 13: What about test_bort_8 and test_bort_d?

test/test_bort.F90
```diff
@@ -0,0 +1,35 @@
+! This is a test for NCEPLIBS-bufr library.
+!
+! This tests the bort() calls of openbf().
+!
```
jbathegit, line 5: How exactly is this tied to openbf()?

test/run_test_bort.sh
edwardhartnett: Next iteration I will add them.

test/test_bort.F90
edwardhartnett: Wrong comment. I will fix.

jbathegit:
Maybe I'm just being dense here, but I don't see how this is configurable to test different bort and bort2 calls in different subprograms.

edwardhartnett:
Next iteration I will demonstrate how this will test other bort() calls.

edwardhartnett:
@jbathegit if we can get this merged I've got the next PR ready to go. ;-)

jbathegit:
I thought you were going to add in test_bort_8 and test_bort_d examples as well.  And by "next iteration" I thought you meant the next follow-on commit for this PR, not an entirely new PR.  Why am I merging this one if it's incomplete and doesn't really do anything?

FWIW, I peeked ahead at your next PR.  Am I correct in understanding that you're not really testing specific errors or specific bort cases in each subprogram, but rather just throwing a bunch of random junk at each subprogram to confirm that it breaks somewhere, perhaps not even in that particular subprogram but maybe somewhere farther down in one of the subprograms that it in turn calls?

I thought the goal here was to test out all of the specific individual bort and bort2 cases inside of each specific subprogram.  But it looks like you're just checking cases where empty or random input is provided.  And in some cases (e.g. writlc) you're not even passing it the correct data types or parameter counts, so wouldn't that generate build errors now that we've got -Werror turned on?

jbathegit:
I'm adding @jack-woollen to this conversation so that he can chime in too with his thoughts.

edwardhartnett:
This PR is an improvement on the current state of play because some testing is taking place that has not taken place before. That is why it should be merged. We don't wait for perfection, but just better.

The next PR does not just throw random stuff at the functions, but is also not complete yet. ;-)

jbathegit:
OK, I'll go ahead and merge this for now, and I'll trust you to finish everything else up in #379.

I guess I was expecting this test code sprint to produce a lot of PRs similar to my new #380 that I just pushed up.

jbathegit (approved):

PR #379, opened 2023-03-13
Testing more borts.

edwardhartnett:
Part of #381.

In this PR I add testing for borts of more functions.

### DIFF
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -31,6 +31,10 @@
 ./test_bort_4 writdx 1
 [ $? != 1 ] &&  exit 1
 
+# Check writsa().
+./test_bort_4 writdx 1
+[ $? != 1 ] &&  exit 1
+
 # If we made it here, all error codes were correctly returned, and the
 # test passed!
 exit 0
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -28,7 +28,7 @@
 [ $? != 1 ] &&  exit 1
 
 # Check writdx().
-./test_bort_4 writdx 4
+./test_bort_4 writdx 1
 [ $? != 1 ] &&  exit 1
 
 # If we made it here, all error codes were correctly returned, and the
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -27,6 +27,10 @@
 ./test_bort_4 wtstat 4
 [ $? != 1 ] &&  exit 1
 
+# Check writdx().
+./test_bort_4 writdx 4
+[ $? != 1 ] &&  exit 1
+
 # If we made it here, all error codes were correctly returned, and the
 # test passed!
 exit 0
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -5,16 +5,26 @@
 #
 # Ed Hartnett 3/12/23
 
+# Check adn30().
+./test_bort_4 adn30 1
+[ $? != 1 ] &&  exit 1
+./test_bort_4 adn30 2
+[ $? != 1 ] &&  exit 1
+
 # Check bort().
 ./test_bort_4 bort 1
 [ $? != 1 ] &&  exit 1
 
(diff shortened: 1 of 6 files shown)


edwardhartnett:
More bort() testing will come, but let's get these working tests into the repo.

The test_bort.F90 program can be used to test any bort situation, but so far I have been adding tests for the low-hainging fruit, borts that can be easily achieved by passing bad input to a subroutine, without any setup required.

In other cases, setup may be required to test a bort.

In some cases, like a bort in closbf() for example, it may be practically impossible to test without causing a memory leak. If our test program opens a file, and then causes a bort on the close call, the file never gets closed and the resources never released, leading to a memory leak, which will cause the test to fail. ;-)

Anyway, I will harvest the low-hanging fruit and we'll see where we are. It may not be possible to test all borts.

test/test_bort.F90
```diff
@@ -6,9 +6,12 @@
 ! Ed Hartnett 3/12/23
 program test_bort
   implicit none
+  character*2 char_short
+  character*30 char_30
 
   integer :: num_args, len, status
   character(len=32) :: sub_name, test_case
+  CHARACTER ADN30
```
jbathegit, line 15: This should be declared as `character*6 adn30` or `character(len=6) adn30`, because it will return up to a 6-byte character string.

test/test_bort.F90
```diff
@@ -25,12 +28,57 @@ program test_bort
...
+        call cmpmsg('W')
+     endif
+  elseif (sub_name .eq. 'writdx') then
+     if (test_case .eq. '1') then
+        call writdx(0, 0, 0)
+     endif
+  elseif (sub_name .eq. 'writlc') then
+     if (test_case .eq. '1') then
+        call writlc(0, 0, 0, 0)
+     endif
```
jbathegit, line 60: writlc has only 3 call arguments, and all but the first one are character strings, not integers.

test/test_bort.F90
```diff
@@ -25,12 +28,57 @@ program test_bort
...
+  elseif (sub_name .eq. 'writsa') then
+     if (test_case .eq. '1') then
+        call writsa(0, 0, 0, 0)
+     endif
+  elseif (sub_name .eq. 'writsb') then
+     if (test_case .eq. '1') then
+        call writsb(0, 0, 0, 0)
+     endif
```
jbathegit, line 68: writsb has only one call argument, not 4.  Or is that the part of what you're testing here - to see how it behaves with an incorrect number of call arguments?

test/run_test_bort.sh
```diff
@@ -5,12 +5,54 @@
...
+
+# Check writsa().
+./test_bort_4 writsa 1
+[ $? != 1 ] &&  exit 1
+
+# Check writsb().
+./test_bort_4 writsb 1
+[ $? != 1 ] &&  exit 1
+
 # If we made it here, all error codes were correctly returned, and the
```
jbathegit, line 56: What about tests for test_bort_8 and test_bort_d?

test/test_bort.F90
edwardhartnett: True, and this is one of those places where test code is not as good as production code.

If I fix this, it gives me a warning on the test. I need the test to compile warning free, and I need to test this bort(), so I'll leave this as character adn30.

test/test_bort.F90
edwardhartnett: Fixed.

test/run_test_bort.sh
edwardhartnett: Still coming. But since bort only accepts char strings anyway, this does not seem urgent. However we'll get there by the end of the week...

test/test_bort.F90
edwardhartnett: No I'm trying to get the arguments correct type and number, but provide a value that causes a bort. I will fix this.

jbathegit (approved):

PR #380, opened 2023-03-14
add new intest10.F90

### LINKED ISSUE #343: add test for openbt.f
This subroutine is currently untested.

It says:
```
C> This subroutine is called as a last resort from within subroutine
C> cktaba(), in the event the latter subroutine is unable to locate
C> a DX BUFR table for use in reading/decoding a given BUFR message.

```
So we need to call it and then find a way to check whether it is set correctly.

### LINKED ISSUE #346: complete testing for arallocf.f
I don't understand what the untested code is doing. Is this error handling?

### LINKED ISSUE #363: complete testing for ufbmex.f

### LINKED ISSUE #364: complete testing of ufbmem.f
This contains large sections of untested code:

```
C  EMERGENCY ROOM TREATMENT FOR MAXMSG ARRAY OVERFLOW
C  --------------------------------------------------

      IF(IPRT.GE.0) THEN
      CALL ERRWRT('+++++++++++++++++++++WARNING+++++++++++++++++++++++')
      WRITE ( UNIT=ERRSTR, FMT='(A,A,I8,A)' )
     . 'BUFRLIB: UFBMEM - THE NO. OF MESSAGES REQUIRED TO STORE ',
     . 'ALL MESSAGES INTERNALLY EXCEEDS MAXIMUM (', MAXMSG,
     . ') - INCOMPLETE READ'
      CALL ERRWRT(ERRSTR)
      WRITE ( UNIT=ERRSTR, FMT='(A,I8,A,I8,A)' )
     . '>>>UFBMEM STORED ', MSGP(0), ' MESSAGES OUT OF ', NMSG, '<<<'
      CALL ERRWRT(ERRSTR)
      WRITE ( UNIT=ERRSTR, FMT='(A,I8,A,I8,A)' )
     . '>>>UFBMEM STORED ', MLAST0, ' BYTES OUT OF ', MLAST, '<<<'
      CALL ERRWRT(ERRSTR)
      CALL ERRWRT('+++++++++++++++++++++WARNING+++++++++++++++++++++++')
      CALL ERRWRT(' ')
      ENDIF
      MLAST=MLAST0
      ENDIF

      IF(IFLG.EQ.2) THEN

C  EMERGENCY ROOM TREATMENT FOR MAXMEM ARRAY OVERFLOW
C  --------------------------------------------------

      IF(IPRT.GE.0) THEN
      CALL ERRWRT('+++++++++++++++++++++WARNING+++++++++++++++++++++++')
      WRITE ( UNIT=ERRSTR, FMT='(A,A,I8,A)' )
     . 'BUFRLIB: UFBMEM - THE NO. OF BYTES REQUIRED TO STORE ',
     . 'ALL MESSAGES INTERNALLY EXCEEDS MAXIMUM (', MAXMEM,
     . ') - INCOMPLETE READ'
      CALL ERRWRT(ERRSTR)
      WRITE ( UNIT=ERRSTR, FMT='(A,I8,A,I8,A)' )
     . '>>>UFBMEM STORED ', MLAST0, ' BYTES OUT OF ', MLAST, '<<<'
      CALL ERRWRT(ERRSTR)
      WRITE ( UNIT=ERRSTR, FMT='(A,I8,A,I8,A)' )
     . '>>>UFBMEM STORED ', MSGP(0), ' MESSAGES OUT OF ', NMSG, '<<<'
      CALL ERRWRT(ERRSTR)
      CALL ERRWRT('+++++++++++++++++++++WARNING+++++++++++++++++++++++')
      CALL ERRWRT(' ')
      ENDIF
      MLAST=MLAST0
      ENDIF

      IF(IRET.EQ.0) THEN
         CALL CLOSBF(LUNIT)
      ELSE
         IF(MUNIT.NE.0) CALL CLOSBF(LUNIT)
         IF(MUNIT.EQ.0) MUNIT = LUNIT
      ENDIF
      IUNIT = MUNIT

```

This looks pretty complex. @jbathegit you might want to look at this one...

jbathegit:
Fixes #343
Fixes #346
Fixes #363
Fixes #364

### DIFF
--- a/src/openbf.f
+++ b/src/openbf.f
@@ -169,18 +169,19 @@ C     messages)
 
       IF(IO.EQ.'QUIET') THEN
 c  .... override previous IPRT value (printout indicator)
-         IF(LUNDX.LT.-1)  LUNDX = -1
-         IF(LUNDX.GT. 2)  LUNDX =  2
-         IF(LUNDX.GE.0) THEN
+         IPRTPRV = IPRT
+         IPRT = LUNDX
+         IF(IPRT.LT.-1) IPRT = -1
+         IF(IPRT.GT. 2) IPRT =  2
+         IF(IPRT.GE.0) THEN
       CALL ERRWRT('++++++++++++++BUFR ARCHIVE LIBRARY+++++++++++++++++')
       WRITE ( UNIT=ERRSTR, FMT='(A,I3,A,A,I3,A)' )
      . 'BUFRLIB: OPENBF - DEGREE OF MESSAGE PRINT INDICATOR '//
-     . 'CHNGED FROM',IPRT,CPRINT(IPRT+1),' TO',LUNDX,CPRINT(LUNDX+1)
+     . 'CHNGED FROM',IPRTPRV,CPRINT(IPRTPRV+1),' TO',IPRT,CPRINT(IPRT+1)
       CALL ERRWRT(ERRSTR)
       CALL ERRWRT('++++++++++++++BUFR ARCHIVE LIBRARY+++++++++++++++++')
       CALL ERRWRT(' ')
          ENDIF
-         IPRT = LUNDX
       ENDIF
 
       IF(IFOPBF.EQ.0) THEN
--- a/test/intest10.F90
+++ b/test/intest10.F90
@@ -1,0 +1,91 @@
+! This is a test for NCEPLIBS-bufr.
+!
+! Reads test file 'testfiles/IN_10' to test ERRWRT branches in ARALLOCF, UFBMEM, UFBMEX, and OPENBT.
+!
+! J. Ator, 3/13/2023
+
+module Share_errstr
+  ! This module is needed in order to share information between the test program and subroutine errwrt, because
+  ! the latter is not called by the former but rather is called directly from within the BUFRLIB software.
+  
+  character*4000 errstr
+  
+  integer errstr_len
+end module Share_errstr
+
+subroutine errwrt(str)
+  ! This subroutine supersedes the subroutine of the same name within the BUFRLIB software, so that we can
+  ! easily test the generation of error messages from within the library.
+  
+  use Share_errstr
+  
+  character*(*) str
+  
+  integer str_len
+  
+  str_len = len(str)
+  errstr ( errstr_len + 1 : errstr_len + str_len + 1 ) = str
+  errstr_len = errstr_len + str_len
+  
+  return
+end subroutine errwrt
+
+program intest10
+  use Share_errstr
+
+  implicit none
+
+  integer*4 isetprm
+
+  integer icnt, iunt, imesg(150), idate, iret
+
+  character cmgtag*8
+
+  print *, 'Testing reading IN_10 to test ERRWRT branches in ARALLOCF, UFBMEM, UFBMEX, and OPENBT'
+
+#ifdef KIND_8
+  call setim8b ( .true. )
+#endif
+
+  if ( ( isetprm ( 'MAXMSG', 125 ) .ne. 0 ) .or. ( isetprm ( 'MAXMEM', 125000 ) .ne. 0 ) ) stop 1
+
+  ! Test some various out-of-bounds verbosity settings, and test the errwrt branch in arallocf.
+  errstr_len = 0
+  call openbf ( 21, 'QUIET', 3 )
+  if ( index( errstr(1:errstr_len), 'ARRAYS WILL BE DYNAMICALLY ALLOCATED USING THE FOLLOWING VALUES' ) .eq. 0 ) stop 2
+  call openbf ( 21, 'QUIET', -2 )
+  call openbf ( 21, 'QUIET', 1 )
+
+  ! Test the errwrt branches in ufbmem.
+  open ( unit = 21, file = 'testfiles/IN_10_infile1', form = 'unformatted')
+  open ( unit = 22, file = 'testfiles/IN_10_infile2', form = 'unformatted')
+  errstr_len = 0
+  call ufbmem ( 21, 0, icnt, iunt )
+  if ( ( icnt .ne. 125 ) .or. & 
+      ( index( errstr(1:errstr_len), 'UFBMEM - THE NO. OF MESSAGES REQUIRED TO STORE ALL MESSAGES' ) .eq. 0 ) ) stop 3
+  call ufbmem ( 22, 0, icnt, iunt )
+  if ( ( icnt .ne. 97 ) .or. & 
+      ( index( errstr(1:errstr_len), 'UFBMEM - THE NO. OF BYTES REQUIRED TO STORE ALL MESSAGES' ) .eq. 0 ) ) stop 4
+
+  ! Reset the input files.
+  call closbf ( 21 )
+  call closbf ( 22 )
+  open ( unit = 21, file = 'testfiles/IN_10_infile1', form = 'unformatted')
+  open ( unit = 22, file = 'testfiles/IN_10_infile2', form = 'unformatted')
+
+  ! Test the errwrt branches in ufbmex.
+  errstr_len = 0
+  call ufbmex ( 21, 21, 0, icnt, imesg )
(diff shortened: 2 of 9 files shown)


jbathegit:
This is still a draft b/c it needs an update to `test/CMakeLists.txt` to run the new intest10.F90 code, which in turn also needs two new `testfiles/IN_10_infile[12]` files to get added to the bufr-11.6.0.tgz archive on emcrzdm.

Given that we're in a group test sprint this week, I figured there would be more testfiles that need to get added to the emcrzdm archive at the end of the week anyway, so I'll just wait to add them all at the same time when I do all of the weekly merges, and that way I can just update all of the key numbers in all of the .yml files once per week rather than multiple times per week.

src/openbf.f
```diff
@@ -169,18 +169,19 @@ RECURSIVE SUBROUTINE OPENBF(LUNIT,IO,LUNDX)
 
       IF(IO.EQ.'QUIET') THEN
 c  .... override previous IPRT value (printout indicator)
-         IF(LUNDX.LT.-1)  LUNDX = -1
-         IF(LUNDX.GT. 2)  LUNDX =  2
-         IF(LUNDX.GE.0) THEN
+         IPRTPRV = IPRT
+         IPRT = LUNDX
```
edwardhartnett, line 173: Seems like you are adding or changing code, not just addding a test. Is this intentional?

test/intest10.F90
```diff
@@ -0,0 +1,91 @@
...
+  call setim8b ( .true. )
+#endif
+
+  if ( ( isetprm ( 'MAXMSG', 125 ) .ne. 0 ) .or. ( isetprm ( 'MAXMEM', 125000 ) .ne. 0 ) ) stop 1
+
+  ! Test some various out-of-bounds verbosity settings, and test the errwrt branch in arallocf.
+  errstr_len = 0
+  call openbf ( 21, 'QUIET', 3 )
+  if ( index( errstr(1:errstr_len), 'ARRAYS WILL BE DYNAMICALLY ALLOCATED USING THE FOLLOWING VALUES' ) .eq. 0 ) stop 2
+  call openbf ( 21, 'QUIET', -2 )
```
edwardhartnett, line 59: Do these calls the openbf() succeed? Seems like you are not checking anything here...

test/intest10.F90
```diff
@@ -0,0 +1,91 @@
...
+  ! Test the errwrt branches in ufbmem.
+  open ( unit = 21, file = 'testfiles/IN_10_infile1', form = 'unformatted')
+  open ( unit = 22, file = 'testfiles/IN_10_infile2', form = 'unformatted')
+  errstr_len = 0
+  call ufbmem ( 21, 0, icnt, iunt )
+  if ( ( icnt .ne. 125 ) .or. & 
+      ( index( errstr(1:errstr_len), 'UFBMEM - THE NO. OF MESSAGES REQUIRED TO STORE ALL MESSAGES' ) .eq. 0 ) ) stop 3
+  call ufbmem ( 22, 0, icnt, iunt )
+  if ( ( icnt .ne. 97 ) .or. & 
+      ( index( errstr(1:errstr_len), 'UFBMEM - THE NO. OF BYTES REQUIRED TO STORE ALL MESSAGES' ) .eq. 0 ) ) stop 4
```
edwardhartnett: Fine to test error strings but I don't intend to do that for bort() testing.

I will just test that the bort() occurs as expected and ignore the error message.

When we make the changes that allow subroutines to return an error code instead of aborting, I will start checking for the correctlness of the error codes too.

test/intest10.F90
```diff
@@ -0,0 +1,91 @@
...
+
+  ! Reset the input files.
+  call closbf ( 21 )
+  call closbf ( 22 )
+  open ( unit = 21, file = 'testfiles/IN_10_infile1', form = 'unformatted')
+  open ( unit = 22, file = 'testfiles/IN_10_infile2', form = 'unformatted')
```
edwardhartnett: Always use the err parameter on opens and check them afterwards to make sure they succeeded,

test/intest7.F90
```diff
@@ -144,7 +144,7 @@ program intest7
   if ( ( nsub .ne. 402 ) .or. ( ibfms ( r8val ) .ne. 1 ) ) stop 18
 
   ! Test the error handling inside of VALX.
-  errstr_len = 1
+  errstr_len = 0
```
edwardhartnett, line 147: What is going on here? How come you''re changing the errstr_len to 0?

test/intest10.F90
jbathegit: What's happening here is that when OPENBF is called with the second argument set to 'QUIET', all it does is dynamically allocate memory (without actually opening a file to the library) and reset the verbosity level (which I needed in order to do in order to verify that the previously-untested code was now being exercised).  So that's why only the one index check after the first call, then the second call to OPENBF is to test the aforementioned code line with LUNDX < -1, and then the third one is just to set it back to a more reasonable verbosity level of 1, which is needed so that I can check on some other errwrt strings later on in the test code.

BTW, the dynamic allocation only happens on the first call to OPENBF.  The library is smart enough to know if memory has already been allocated and won't do it again during any subsequent calls to OPENBF.

src/openbf.f
jbathegit: Good catch, and yes this is intentional.  Within the new test code, I took the opportunity to test out lines 172 and 173 to make sure they worked when LUNDX was < -1 or > 2.  I did this by passing in LUNDX as a hardcoded constant rather than a variable set to a value, which is not an uncommon way for OPENBF to be called, and that's when it segfaulted trying to (re)set the value of a constant.  So yay for testing ;-) and that's why I made these changes.

test/intest10.F90
jbathegit: OK

test/intest7.F90
jbathegit: Good question!  When I was copying the errwrt subroutine and Share_errstr module over from one of my other test codes, I realized that errstr_len should be 0 instead of 1, because the logic in the errwrt subroutine fills the errstr array starting with byte errstr_len+1, rather than starting with byte errstr_len.  So it hadn't been causing any problems, but it also was never using the first byte of the array and was always filling it beginning with byte 2.  So I went ahead and fixed it in the other test codes as well.

test/intest10.F90
jbathegit: How about if I use iostat= instead of err= and then just check the value of the associated variable?  If I use err= then I have to give it a statement label to branch to in case of a problem.

edwardhartnett: Yes, totally cool. Just test every single statement if possible. If an error code is returned by any operation, get it and check it and stop if it's not what you want it to be.

src/openbf.f
edwardhartnett: If I understand you correctly you've fixed the code so that the segfault cannot occur with bad input to the function - is that correct?

If so, did you also add testing for that?

test/intest10.F90
edwardhartnett: Can you please put this explanation into the code as comments? It's far from obvious!

jbathegit: OK, I'll try to expand the existing comment to make it clearer.

src/openbf.f
jbathegit: Yes, that's correct, and yes I'm testing for that with the following two lines of intest10.F90 which test these out-of-bounds conditions:

```
  call openbf ( 21, 'QUIET', 3 )
  call openbf ( 21, 'QUIET', -2 )
```

As noted earlier, this is how I found the error in the first place and can now verify that the fix I made is working! ;-)

edwardhartnett:
BTW I didn't intend that we would wait until the end of the week to merge tests, just that we would get together and catch up on any open PRs at the end of the week. I intend to submit daily PRs, and would like to see them reviewed and merged as the week progresses. I suggest you do the same, even if you have to update the data files multiple times.

jbathegit:
OK, I'll get this pushed up sooner rather than later ;-)

edwardhartnett (approved):

Issue #381, opened 2023-03-14
test all aborts in the code which are caused by bad input parameters


edwardhartnett:
We have a lot of cases where passing a bad parameter to a subroutine will call a bort(). These all need to be tested.

We are doing this testing with program test_bort.F90 and test script run_test_bort.sh.

edwardhartnett:
Great progress here, only a few borts remain untested...

PR #383, opened 2023-03-14
Testing more borts 2

edwardhartnett:
Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/381.

In this PR I add testing for borts of more functions.

### DIFF
--- a/src/ufbcnt.f
+++ b/src/ufbcnt.f
@@ -1,6 +1,6 @@
 C> @file
 C> @brief Get the current message number and data subset number within
-C> a BUFR file
+C> a BUFR file.
 C>
 C> @author J. Woollen @date 1994-01-06
 
@@ -8,15 +8,6 @@ C> This subroutine returns the current location of the file pointer
 C> within a BUFR file, in terms of a message number counting from the
 C> beginning of the file, and a data subset number counting from the
 C> beginning of that message.
-C>
-C> @param[in] LUNIT -- integer: Fortran logical unit number for BUFR file
-C> @param[out] KMSG -- integer: Ordinal number of current message,
-C>                     counting from the beginning of the BUFR file, but
-C>                     not counting any messages which contain DX BUFR
-C>                     tables information
-C> @param[out] KSUB -- integer: Ordinal number of current data subset
-C>                     within (KMSG)th message, counting from the
-C>                     beginning of the message
 C>
 C> @remarks
 C> - Logical unit LUNIT should have already been opened via a previous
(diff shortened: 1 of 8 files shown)


src/ufbcnt.f
```diff
@@ -33,6 +24,15 @@
...
+C> @param[in] LUNIT -- integer: Fortran logical unit number for BUFR file
+C> @param[out] KMSG -- integer: Ordinal number of current message,
+C>                     counting from the beginning of the BUFR file, but
+C>                     not counting any messages which contain DX BUFR
+C>                     tables information
+C> @param[out] KSUB -- integer: Ordinal number of current data subset
+C>                     within (KMSG)th message, counting from the
+C>                     beginning of the message
+C>
 C> @author J. Woollen @date 1994-01-06
```
jbathegit, line 36: Just curious, why are you moving the parameter descriptions in all of these routines to be _after_ all of the remarks and discussion?  The remarks and discussion notes refer to some of the parameters listed in the parameter descriptions, so it seems more sensible to keep those remarks and notes after the parameter descriptions like they were previously; otherwise, they're talking about concepts that haven't even been introduced yet!

test/run_test_bort.sh
```diff
@@ -27,6 +27,108 @@
 ./test_bort_4 cmpmsg 1
 [ $? != 1 ] &&  exit 1
 
+# Check codflg().
+./test_bort_4 codflg 1
+[ $? != 1 ] &&  exit 1
+
```
jbathegit, line 33: Shouldn't this run_test_bort.sh script be in the test_scripts subdirectory along with all of the other test scripts, rather than in the test directory with all of the source code?

src/ufbcnt.f
edwardhartnett: That's the way it is done in all the NCEPLIBS and I'm trying to maintain a similar look and feel everywhere. Description, parameters, return codes, author is how we do it everywhere else.

test/run_test_bort.sh
edwardhartnett: Well that is open to discussion.

All the scripts in test_scripts are for utils, and I was planning on moving all of them to a test_utils directory.

jbathegit: Actually, the test_outtest.sh script (wrapper script for all of the outtest*F90 codes) is also in test_scripts, so it's not just scripts for testing utilities in that directory.

That said, if you wanted to have a separate test_utils directory for all of the utility scripts, and then a separate test_scripts directory containing just your new test_bort.sh script and the test_outtest.sh script, I wouldn't be opposed to that ;-)

jbathegit (approved):

PR #385, opened 2023-03-14
Complete testing for getlens.f

### LINKED ISSUE #351: complete testing of getlens.f
The code says:

```
C> This subroutine reads the lengths of all of the individual
C> sections of a given BUFR message, up to a specified point in
C> the message.

```

So it seems like a good idea to use one of the test input files, and make sure this function comes up with the correct answer for the test file.

AlexanderRichert-NOAA:
Fixed #351

This PR adds testing for the rest of getlens.f.

### DIFF
--- a/test/intest1.F90
+++ b/test/intest1.F90
@@ -102,7 +102,7 @@ program intest1
   IF (softvstg(5:12) .ne. '5.8.5.10') stop 17
 
   call getlens(ibfmg,3,l0,l1,l2,l3,l4,l5)
-  if (.not. all((/l0, l1, l2, l3, l4, l5/) .eq. (/8,22,0,24,-1,-1/))) stop 18
+  if (.not. all((/l0,l1,l2,l3,l4,l5/) .eq. (/8,22,0,24,-1,-1/))) stop 18
 
   ! Close the test file.
   call ccbfl()
--- a/test/intest1.F90
+++ b/test/intest1.F90
@@ -8,7 +8,7 @@ program intest1
   implicit none
   integer mxbfd4, mxds3, nds3, ierme, imgdt
   integer ierndv, iernds, mxr8pm, mxr8lv, iertgp, nr8lv
-  integer l0, l1, l2, l3, l4, l5, l6
+  integer l0, l1, l2, l3, l4, l5
   integer*4 mxbf, nbyt, ierr
   integer*4 iupbs01, iupbs3, ireadsb, ibfms
   parameter (mxbf = 20000)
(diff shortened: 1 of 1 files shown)


test/intest1.F90
```diff
@@ -100,6 +101,9 @@ program intest1
   call readlc(11, softvstg, 'SOFTV')
   IF (softvstg(5:12) .ne. '5.8.5.10') stop 17
 
+  call getlens(ibfmg,3,l0,l1,l2,l3,l4,l5)
```
edwardhartnett: Can you give me a one-line comment as to what is happening?

AlexanderRichert-NOAA: `getlens` returns the lengths of BUFR sections in the `ibmfg` bufr message (up to and including section 3, in this case). Because `setim8b` has been invoked above, the `IF(IM8B)` clause in `getlens` is triggered so it returns 8-byte integers for the section lengths.

edwardhartnett: Yes, but add that explanation to the code as a comment. ;-)

jbathegit: Any chance you could modify the test to check all the way through Section 5 so that **all** of the lines of getlens get tested, and not just the ones that do Sections 0 through 3? ;-)

Also, why do you have variable l6 declared?  Unless I'm missing something, it looks like it's unused.

Finally, I would suggest to take the "(returns 8-byte ints for KIND_8)" out of the comment, because that's no different than for any other subroutine call as well.  In other words, just "Get and check lengths of Sections 0-3 (or 5 ;-)" should be enough.

test/intest1.F90
```diff
@@ -8,6 +8,7 @@ program intest1
   implicit none
   integer mxbfd4, mxds3, nds3, ierme, imgdt
   integer ierndv, iernds, mxr8pm, mxr8lv, iertgp, nr8lv
+  integer l0, l1, l2, l3, l4, l5
   integer*4 mxbf, nbyt, ierr
   integer*4 iupbs01, iupbs3, ireadsb, ibfms
```
jbathegit, line 13: Thanks Alex, and just one last request - can we change the variable names to "len0", "len1",...,"len5", rather than "l0, l1,...,l5"?  Sorry, but every time I look at this I'm reading integers 10, 11,...,15, b/c the 1 looks too much like the lowercase l.

Otherwise, looks good - and thanks!

jbathegit (approved):

edwardhartnett:
Awesome! @AlexanderRichert-NOAA there are a bunch of other test issues where the same problem occurs, that is, the 8-byte int part of the code is not tested. Now that you know how, why don't you grab some of those for next week.

Issue #376, opened 2023-03-13
Simplify util tests


edwardhartnett:
The util tests can be considerably simplified. For example, we have test_binv.sh, which contains:

```
set -eu

cmd=$1
outfile=$2
reffile=$3

rc="-1"
$cmd > $outfile && diff -w $outfile $reffile
rc=${?}

exit $rc

```

This is then used in the cmake file like this:

```
# Test binv utility
add_test(NAME test_binv  
  COMMAND ${CMAKE_BINARY_DIR}/bin/test_binv.sh "${CMAKE_BINARY_DIR}/utils/binv testfiles/data/satwndbufr" "testrun/binv.out"  "testfiles/testoutput/binv.out"
  )

```

This could all be replaced by one line of script code which runs the binv test with the arguments. If any aspects of the shell script need to be set at configure time, then we use test_binv.sh.in.

The cmake file should run the binv test script without providing any test details, those can be in the script file.

We need to add tests to the binv test script. Want the build script to contain all the testing details, for simple debugging. 

It would be good if this could be done soon, so that we can proceed to test all utilities more thoroughly.

jbathegit:
A question, and assuming I'm understanding correctly that your intent here is to have one `add_test()` for each utility which does nothing more than run the script, and where the script itself then does one or more runs of the actual utility for various cases...

Will gcovr still properly track code coverage for all such utilities?  In other words, if multiple runs of the utility are now being done inside of a single script kicked off by a single `add_test()` (vs. each run of the utility being kicked off by its own separate `add_test()`), will gcovr still track everything the same way?

edwardhartnett:
Yes, gcovr somehow figures it all out and still works.

This is how we run tests in netcdf-c, netcdf-fortran, PIO, and many other projects.

edwardhartnett:
I believe this is done so I will close this issue.

jbathegit:
It's not quite done yet.  I just finished the test updates for the remaining utilities (cmpbqm, readbp, and readmp) and will push up a new PR soon.

edwardhartnett:
Press on!

PR #388, opened 2023-03-17
documentation, more testing for adn30()/idn30(), changes in gcovr output

### LINKED ISSUE #389: complete testing for adn30()/idn()
There are some untested parts of these simple subroutines.

edwardhartnett:
Fixes #389

Some documentation fixes, adding tests for adn30()/idn30() to bring them to full testing.

Also some changes to the gcovr run, adding some new parameters, which results in better tracking of branches testes, and a new column, "decisions", which is supposed to better help us ensure that all decisions are fully covered. We'll have to use this a bit to see how it helps.

See https://gcovr.com/en/stable/faq.html for info about the new gcovr options.

### DIFF
--- a/src/adn30.f
+++ b/src/adn30.f
@@ -1,12 +1,11 @@
 C> @file
-C> @brief Convert an FXY value from its bit-wise (integer)
-C> representation to its five or six character representation.
+C> @brief Convert an FXY value from an integer
+C> to a character string of length 5 or 6.
 C>
 C> @author J. Woollen @date 1994-01-06
 
-C> This function converts an FXY value from its bit-wise
-C> (integer) representation to its 5 or 6 character
-C> representation.
+C> This function converts an FXY value from an integer
+C> to a character string of length 5 or 6.
 C>
 C> This function is the logical inverse of function idn30().
 C>
@@ -26,7 +25,6 @@ C> @author J. Woollen @date 1994-01-06
 
 C----------------------------------------------------------------------
 C----------------------------------------------------------------------
-
       IF(LEN(ADN30).LT.L30         ) GOTO 900
       IF(IDN.LT.0 .OR. IDN.GT.65535) GOTO 901
       IF(L30.EQ.5) THEN
--- a/src/arallocc.c
+++ b/src/arallocc.c
@@ -17,15 +17,8 @@
  * dynamically allocate internal C language arrays based on parameter
  * values set during one or more previous calls to function isetprm().
  *
- * This subroutine isn't normally called directly from an application
- * program, since it's automatically called internally during the first
- * call to subroutine openbf() from an application program.
- *
- * @remarks
  * All memory allocated within this subroutine can be freed via a
- * subsequent call to subroutine exitbufr() from within the
- * application program, or else it will be freed automatically by the
- * operating system once the application program terminates.
+ * subsequent call to subroutine exitbufr().
  *
  * @author J. Ator @date 2014-12-04
  */
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -55,7 +55,7 @@ jobs:
       run: |
         cd bufr/build
         ctest --verbose --output-on-failure --rerun-failed
-        gcovr --root .. -v  --html-details --exclude ../test --exclude CMakeFiles --print-summary -o test-coverage.html
+        gcovr --root .. -v  --html-details --exclude ../test --exclude CMakeFiles --print-summary --exclude-unreachable-branches --exclude-throw-branches --decisions -o test-coverage.html
 
     - name: cache-data
       if: steps.cache-data.outputs.cache-hit != 'true'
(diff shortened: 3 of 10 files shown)


src/adn30.f
```diff
@@ -1,12 +1,11 @@
 C> @file
-C> @brief Convert an FXY value from its bit-wise (integer)
-C> representation to its five or six character representation.
+C> @brief Convert an FXY value from an integer
```
edwardhartnett, line 2: @jack-woollen or @jbathegit is FXY an acronym? Or what is it? What does it mean?

src/adn30.f
```diff
@@ -26,7 +25,6 @@ FUNCTION ADN30(IDN,L30)
 
 C----------------------------------------------------------------------
 C----------------------------------------------------------------------
-
       IF(LEN(ADN30).LT.L30         ) GOTO 900
```
edwardhartnett, line 28: What is the significance of the number 30 to this function?

The number 30 is used in the name of the function and in some variable names, but what does this function have to do with 30?

src/adn30.f
jbathegit: FXY isn't an acronym.  It's a reference notation that's used in the official (WMO) BUFR regulations for Section 3 descriptors, which are normally 6 digits in the form of FXXYYY, where F ranges between 0-3, XX ranges between 0-63, and YYY ranges between 0-255.

Bottom line, the term is well-defined in the official manual.

src/adn30.f
jbathegit: That's a really good question!

@jack-woollen, does that number have some sort of legacy significance back from the early days of library development?  If not, and if there's no other reason to keep it, then maybe we should remove it from this and the idn30 routines to prevent further confusion.

src/arallocc.c
```diff
@@ -17,15 +17,8 @@
...
- *
- * @remarks
  * All memory allocated within this subroutine can be freed via a
- * subsequent call to subroutine exitbufr() from within the
- * application program, or else it will be freed automatically by the
- * operating system once the application program terminates.
+ * subsequent call to subroutine exitbufr().
  *
  * @author J. Ator @date 2014-12-04
  */
```
jbathegit, line 24: The last note, about allocated memory being freed automatically once a program terminates, might seem obvious to serious programmers.  But I also didn't want every user to think they now had to add an explicit call to exitbufr to their application code, and which is why I put that note in there in the first place back when I added the dynamic (run-time) allocation capability to the library back in 2014.

The idea behind exitbufr was that an application code would only ever need to explicitly call it if for some reason they either:

- needed to reallocate memory to resize internal library arrays (e.g. if they suddenly needed them to be larger), or
- if they knew for sure that they were completely done with any and all use of the library and wanted to free up all of the allocated memory before moving on to other memory-intensive tasks within their code

But this is clearly noted in the remarks within exitbufr, so I'm OK if you now want to simplify the remarks in arallocc and remove it here.  Though if you do so, I would also ask that you make the same updates to the docblock in arallocf, which has the same content.

src/adn30.f
edwardhartnett: Well we can't remove the 30 without changing the function names, which is backwards incompatible.

But we could document the reason for 30.

src/arallocc.c
edwardhartnett: In fact, users should always free any memory that they allocated.

If not, then their software will fail memory testing. And also any package that uses their software will fail memory testing.

Leaks are serious problems, especially on HPC systems where the leak may be replicated on 10000 processors.

I believe we maintain a count of open files within the library, right? When that count reaches 0, we should call this function. Then the user will be out of the loop.

src/adn30.f
jbathegit: No application code should be calling these functions directly, so the only issue should be just making all of the changes internal to the library itself and within our test_bort code so that it's internally compatible.

We should still document this change in the release notes, but I don't really see backwards-incompatibility as an issue, since no external codes should be directly referencing either of these functions.

But of course I would like to hear from Jack in case there's some reason for the 30 that we're not aware of?

src/arallocc.c
jbathegit: Only one of our test codes (outtest3) is currently calling exitbufr, just to exercise it for gcovr.  None of our other test codes call it, nor do any of our utilities for that matter, yet none of them are failing memory testing.

Am I missing something here?  I thought the operating system always freed any allocated memory by default once the process terminated.

edwardhartnett: That is a good question. I don't know the answer but I will look into it.

edwardhartnett:
Can we merge this before our meeting this afternoon?

jbathegit:
OK, I can do that, but first could you please push up another commit to make the remarks in the arallocf docblock the same as the changes you've now made to the remarks in the arallocc docblock?  I think these should stay consistent, and I asked about this earlier, but I can certainly understand if it got lost in the shuffle of all of the other discussion we've been having ;-)

Once that's done, I'll approve and merge it, and then I'll be following right behind with a PR of my own that's now ready for you to look at for a number of the utility codes in #376 ;-)  The timing will be good b/c I have changes of my own to test/CMakeLists.txt that I can merge on top of yours.

jbathegit (approved):

PR #390, opened 2023-03-17
expanded README description of projects that use NCEPLIBS-bufr

edwardhartnett:
I have explanded the explanation of each project listed in the README that uses NCEPLIBS-bufr.

Does anyone know who or what uses the bufr-dump utility? Is it part of any systems?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ NCEPLIBS-bufr is used by numerous other projects including:
   Operational Model Archive and Distribution System (NOMADS)](https://nomads.ncep.noaa.gov/).
 - [prepobs](https://github.com/NOAA-EMC/prepobs) from the [The NCEP Production Suite](https://www.nco.ncep.noaa.gov/pmb/prod_overview/).
 - [bufr-dump](https://github.com/NOAA-EMC/bufr-dump).
-- [the Global Ensemble Forecast
+- the [Global Ensemble Forecast
   System(GEFS)](https://www.ncei.noaa.gov/products/weather-climate-models/global-ensemble-forecast).
-- [the High-Resolution Rapid Refresh model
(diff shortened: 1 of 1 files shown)


jbathegit:
bufr_dump is run by all of the model DA systems when it's time to collect data for use in the analyses.  We have an observational database consisting of what we call "tanks" and which are basically a continuous feed of any number of types of data that we receive from the outside world and encode into BUFR for storage in our database.  Think of it like a bunch of buckets being continually filled with new obs as they come in each day, and stored by type and date.

So, when it's time to run one of the models, what bufr_dump does is go into that database and pull out just the observations that are needed for that model, and still in BUFR format.  bufr_dump is given a bunch of parameters to tailor the request, such as which data type(s) are needed, and within what time window, and within what lat/long box in case of a regional or non-global domain model, etc., and bufr_dump goes and fills that request and presents the corresponding BUFR observations back to the user, while at the same time also filtering out duplicates, applying rudimentary QC including any specific inclusions or exclusions requested by the SDM, etc.

In fact, bufr_dump can also be run directly by users for extracting certain types of data from the database for their own testing, research, etc., so it's not just used by operational assimilation processes.

Glad you asked? ;-)

BTW, and while you're editing the README file anyway, do you want your name under Author to read as "Edward Hartnett" or "Ed Hartnett"?

edwardhartnett:
Thanks for that explanation. I think it would be good to submit some version of that as an update to the bufr_dump GitHub page. It's not at all obvious what it does!

"Edward Hartnett" is my official name so I might as well use it. ;-)

jbathegit:
OK, your call.  "Jeffrey" is my official name, but I just prefer to be referred to as "Jeff", so that's why I asked ;-)

jbathegit (approved):

Issue #397, opened 2023-03-18
improve documentation


edwardhartnett:
As a result of the testing campaign we are currently undertaking, there will be a lot of documentation improvements.

edwardhartnett:
When we look at the built documentation, we see a lot of lists of things, and their brief descriptions. For example:

![image](https://user-images.githubusercontent.com/38856240/227781600-3167d0e7-3a7a-424c-ab4d-fedf30bfecd9.png)

The problem here is all the wasted space saying: "This module contains ".

To combat this, we will adopt an oft-used doxygen convention, which is to write our brief descriptions as if those words were already present. So instead of:

`!> This module declares and initializes the MAXMEM variable.
`
we write:

`!> Declare and initialize the MAXMEM variable.`

This applies to subroutines/functions as well, where we need to change from:

```
C> This function calls BUFRLIB subroutine readmm() and passes
C> back its return code as the function value.

```
to:

```
C> Calls BUFRLIB subroutine readmm() and pass
C> back its return code as the function value.

```

I will start changing this in every documentation I touch in testing, others should do the same. Remember that the brief description is supposed to be brief. So I would further edit the above to:

```
C> Call readmm() and returns its return code.

```

edwardhartnett:
What exactly do we mean by:

"bit-wise representation of FXY number"?

Isn't this just the same as "number"? That is, every number on the  computer is a bit-wise representation of a number. I see from the documentation of ifxy,f:

C> @remarks
C> Per the [official WMO BUFR regulations](@ref manual), an FXY value
C> can be represented as a bit-wise integer in 16 bits, ordered from
C> left (most significant) to right (least significant), and where the
C> F value occupies the first 2 bits, the X value occupies the next 6
C> bits, and the Y value occupies the last 8 bits.
C>
C>     For example, if ADSC = '063022'
C>
C>        F |     X     |      Y
C>        0 |     63    |      22
C>       0 0 1 1 1 1 1 1 0 0 0 1 0 1 1 0
C>
C>     then the corresponding ifxy
C>
C>       = ( 2**13 + 2**12 + 2**11 + 2**10 + 2**9 + 2**8 +
C>           2**4 + 2**2 + 2**1 )
C>
C>       = 16150

So this makes it a little more clear. This is a WMO thing. So perhaps we should be calling it a "WMO bit-wise representation of FXY number"?

edwardhartnett:
[...]

PR #398, opened 2023-03-18
Documentation improvements

edwardhartnett:
Part of #397

Documentation improvements. No code changes or new tests in this PR, just documentation.

### DIFF
--- a/src/elemdx.f
+++ b/src/elemdx.f
@@ -9,7 +9,7 @@ C> definition card that was previously read from a user-supplied DX BUFR
 C> table file in character format by subroutine rdusdx().
 C> These decoded values are then added to the
 C> already-existing entry for that mnemonic within the internal BUFR
-C> Table B array TABB(*,LUN) in module tababd.
+C> Table B array TABB(*,LUN) in module @ref moda_tababd.
 C>
 C> @param[in] CARD - character*80: mnemonic definition card that was read
 C> from a user-supplied DX BUFR table.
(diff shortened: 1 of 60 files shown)


edwardhartnett:
OK, I have discovered that if we say "@ref moda_tababd" instead of just "tababd", then doxygen generates a nice link to that module.

So for example:

```
C> This subroutine initializes the internal arrays
C> (in module @ref moda_tababd) holding the DX BUFR table. It then

```
The second line generates a link to the documentation for moda_tababd, which is very helpful.

I have made this change for most or all of the existing references to modules in the documentation, but if you see one that is not referenced, or add new documentation referencing modules, then please use this notation to generate the hyperlink in the doxygen output.

src/elemdx.f
edwardhartnett, line 1: In this file I applied emacs autoindent and used other automatic emacs tools to format the comments for line length and whitespace. Only formatting changes for this file.

src/modules_arrs.F90
edwardhartnett, line 1: In this file and in modules_vars.F90 I applied emacs autoindent and used other automatic emacs tools to format the comments for line length and whitespace. Only formatting changes for these files.

jbathegit: OK, but I thought one of the advantages of using F90 was that we could also have comments going out past column 72, and not just the source code going out past column 72.

edwardhartnett: Sure but we don't have to have extra long lines. ;-)

jbathegit:
The @ref feature is a nice touch for the modules, so thanks for figuring that out!

FWIW, I'd only been using @ref previously to reference other points in the documentation that had already been tagged using @anchor.  So this is good to know that it works automatically for defined modules!

jbathegit (approved):

Issue #300, opened 2023-01-24
Achieve and enforce warning-free builds with GNU -Wall -Werror in CI


edwardhartnett:
We can turn on full warning checks in GNU gfortran and then clear all warnings.

Then we can turn on -Werror and ensure that no code with warnings gets committed to the repo.

edwardhartnett:
I am still getting warnings when I build like this:

cmake -DTEST_FILE_DIR=/home/ed/bufr -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/usr/local/NCEPLIBS-sp.2.4.0 -DENABLE_DOCS=ON .. && make VERBOSE=1 && make test

For example:

cd /home/ed/b2/b/src && /usr/bin/gfortran -DLITTLE_ENDIAN -I/home/ed/b2/b/src -I/home/ed/b2/src -g -fbacktrace -ggdb -Wall -O0 -Jinclude/bufr_4 -fPIC -c /home/ed/b2/src/copymg.f -o CMakeFiles/bufr_4_f.dir/copymg.f.o
/home/ed/b2/src/copymg.f:85:72:

   85 |       SUBSET = TAG(INODE(LIN))
      |                                                                        1
Warning: CHARACTER expression will be truncated in assignment (8/10) at (1) [-Wcharacter-truncation]


and

cd /home/ed/b2/b/src && /usr/bin/gfortran -DLITTLE_ENDIAN -I/home/ed/b2/b/src -I/home/ed/b2/src -g -fbacktrace -ggdb -Wall -O0 -Jinclude/bufr_4 -fPIC -c /home/ed/b2/src/cpymem.f -o CMakeFiles/bufr_4_f.dir/cpymem.f.o
/home/ed/b2/src/cpymem.f:83:72:

   83 |       SUBSET = TAG(INODE(LIN))
      |                                                                        1
Warning: CHARACTER expression will be truncated in assignment (8/10) at (1) [-Wcharacter-truncation]


OK, the problem is that in the developer workflow, we are not using -Wall:

[...]

PR #402, opened 2023-03-20
upgraded developer workflow to gcc-11/gfortran-11, also remove unneeded flag  -fallow-invalid-boz

### LINKED ISSUE #403: remove -fallow-invalid-boz flag from build
The -fallow-invalid-boz flag does this:

```
-fallow-invalid-boz
A BOZ literal constant can occur in a limited number of contexts in standard conforming Fortran. This option degrades an error condition to a warning, and allows a BOZ literal constant to appear where the Fortran standard would otherwise prohibit its use.

```

In many other NCEPLIBS we need this flag because we initialize BOZ constants in non-standard ways. But not so NCEPLIBS-bufr, which currently builds and tests just fine without -fallow-invalid-boz. So I will take it out of the build.

edwardhartnett:
Part of #300
Fixes #403

### DIFF
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -45,7 +45,7 @@ jobs:
         cd bufr
         mkdir build
         cd build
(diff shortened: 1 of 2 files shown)


edwardhartnett:
OK, funny thing is that switching to gcc-12/gfortran-12 causes gcovr to break. @GeorgeGayno-NOAA are you running your developer CI with gcc-12/gfortran-12? If so, that could explain the gcovr problem you've seen...

GeorgeGayno-NOAA:
> OK, funny thing is that switching to gcc-12/gfortran-12 causes gcovr to break. @GeorgeGayno-NOAA are you running your developer CI with gcc-12/gfortran-12? If so, that could explain the gcovr problem you've seen...

Looks like we are using mpifort/mpicc: https://github.com/ufs-community/UFS_UTILS/blob/develop/.github/workflows/developer.yml#L28

jbathegit (approved):

PR #406, opened 2023-03-21
update strnum to handle valx functionality, and remove valx

### LINKED ISSUE #369: can we remove or greatly simplify strnum.f?
We have strnum.f. It says:

```
C> This subroutine decodes an integer from a character string.  The
C> string should contain only digits and (optional) trailing blanks.
C> It should not contain any sign ('+' or '-') character nor any
C> leading blanks nor embedded blanks.

```
Can this not be done with standard fortran in a single command? Why do we have a subprogram for this?

### LINKED ISSUE #407: why do we use STRSUC() instead of trim()?
We have:

```
C> This subroutine removes leading and trailing blanks from a
C> character string.  The string may not contain any embedded blanks.
C>
C> @param[in]  STR1 -- character*(*): String
C> @param[out] STR2 -- character*(*): Copy of STR1 with leading and
C>                     trailing blanks removed
C> @param[out] LENS -- integer: Length of STR2
C>                     - -1 = STR1 contained embedded blanks
C>
      SUBROUTINE STRSUC(STR1,STR2,LENS)

```

How is this different from trim()?

jbathegit:
Fixes #369
Fixes #407

Part of #254
Part of #300


jack-woollen (approved):
Good idea to get rid of valx!

src/strnum.F90
```diff
@@ -0,0 +1,73 @@
...
+
+     im8b = .true.
+     return
+  end if
+
+  iret = 0
+
+! Check for blank input string.
+
+  if ( str .eq. ' ' ) then
```
edwardhartnett, line 54: What happens if str is '  '? (That is, two spaces, not one?) 

Is there a test that tries this function with different bad inputs to see if it catches them?

Until now, we have been relying on intestN and outtestN to test lots of functions. These are more like system tests than unit tests. What would be good would be (in addition) a new test, something like test_strings.F90, which can test all the string handling functions of NCEPLIBS-bufr. In that new test, you can systematically check the strnum subprogram with a variety of inputs and check that it catches all error conditions. See test_misc.F90 for an example of a test that runs these sorts of simple unit tests.

src/strnum.F90
```diff
@@ -0,0 +1,73 @@
...
+  call strsuc ( str, str2, lens )
+  if ( verify ( str2(1:lens), "0123456789+-" ) .ne. 0 ) then
+    iret = -1
+    return
+  end if
+
+! Decode the integer from the string.
+
+  write ( fmt, '(''(I'',I2,'')'')' ) lens
+  read ( str2(1:lens), fmt ) num
```
edwardhartnett, line 70: If non-numeric characters are in the string, won't this fail? In which case, why try to detect non-numeric characters? Why not just try to read and if there is a failure, report a failure?

src/strnum.F90
```diff
@@ -0,0 +1,73 @@
...
+! Check for blank input string.
+
+  if ( str .eq. ' ' ) then
+    num = 0
+    return
+  end if
+
+! Verify that the string contains all legal characters.
+
+  call strsuc ( str, str2, lens )
```
edwardhartnett, line 61: Why strsuc? Can't trim() do this already?

src/strnum.F90
jbathegit: This check will work for a string of any number of blanks (not just a single blank).  Within this PR I've already modified intest7.F90 accordingly to confirm this (it tests for a string of 4 blanks ;-).

That said, if you'd prefer I/we can move these new strnum tests over from intest7.F90 to test_misc.F90 - just let me know.

src/strnum.F90
jbathegit: IMHO, the prior `verify` check was a cleaner and clearer way to confirm that str2(1:lens) contained only valid characters, and in such cases to go ahead and report the failure by setting iret to a value of -1.

That said, I certainly could change this to add an err= option to the read statement and have it reroute control to a new numbered statement to set iret = -1 and return that way, if you and/or Jack think that would be a better way to do it.  Again, this was just my personal preference.

src/strnum.F90
jbathegit: If you look at my new strsuc in this PR (where I've completely rewritten it and migrated it to F90), you'll see that it now does nothing more than call the intrinsic functions adjustl and len_trim ;-)

It's true that, after having done that, I could have just done away with strsuc altogether and just called adjustl and len_trim directly from within strnum.F90.  But then I'd have to do likewise in a bunch of different subprograms in the library which also call strsuc, so I decided (again, just IMHO) to do it this way.

src/strnum.F90
jbathegit: OK, I just saw your #409 (sorry, it's hard keeping track of all of these overlapping conversations and PR's flying around ;-)

So I take it you would indeed prefer that I just modify the new strnum.F90 to add an iostat= check to the read and remove the verify check, and I see where you've already included some very thorough strnum testing in test_misc.F90, so I'll go ahead and remove my less thorough tests from intest7.F90 as well.

edwardhartnett:
This is a solid improvement. I suggest you merge this and we continue discussions in #369.

Removing valx() is a great thing.

jbathegit:
OK, I'll do that, and I see you've now closed #409.  But I'll still follow up with another PR of my own which follows your advice in #409, and which cleans up strnum.F90 even further and moves the testing for some of those oddball cases from intest7.F90 to test_misc.F90, like you demonstrated.

I'll do that tomorrow morning...

src/strnum.F90
edwardhartnett: We want to reduce code to the smallest amount that can meet current requirements. This is the lowest cost for NOAA over the long term, and the least amount of work for us.

So if using iostat can save us a few lines of code, those are lines NOAA does not have to pay for. Recall that maintenance costs for a line of code are an order of magnitude more than the cost to write the line of code. Code is not free to keep, it's like owning a horse - it costs a lot of money even if you don't use it much.

We must always prefer the solution that meets current requirements with the least lines of code.

PR #408, opened 2023-03-21
Demonstrate how to replace strnum() with standard fortran internal read()

edwardhartnett:
Part of #369 

@jbathegit take a look at the code in test_misc.F90. Doesn't it do everything that strnum() does, using only standard Fortran internal reads?


edwardhartnett:
Whoops, messed this one up. I'll try again...

PR #409, opened 2023-03-21
using standard fortran internal read to demonstrate how strnum() may be replaced

edwardhartnett:
Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/369

@jbathegit take a look at the code in test_misc.F90. Doesn't it do everything that strnum() does, using only standard Fortran internal reads?

### DIFF
--- a/test/test_misc.F90
+++ b/test/test_misc.F90
@@ -9,15 +9,17 @@ program test_bort
   character*5 char5
   character*5 adn30
   integer a, idn30
+  integer i1, i2, ios
+  character*8 st1, st2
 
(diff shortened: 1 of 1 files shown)


edwardhartnett:
This was not intended to be merged but just to demonstrate that an internal read() can be used to replace the strnum() function.

If we can remove a function from the library, as you have done with valx(), that is a great cost savings for NOAA and time saver for every programmer involved with the code, moving forward.

I will close this PR and we can continue discussion in #369.

PR #411, opened 2023-03-24
atrcpt testing

### LINKED ISSUE #347: complete testing of atrcpt.f
The untested code is the I8 integer code. How can that be? I thought all tests were run on all kinds of the library?

AlexanderRichert-NOAA:
Setting up a draft PR for adding 8-byte testing for atrcpt.

Fixes #347


AlexanderRichert-NOAA:
@jbathegit I could use some guidance here. I think that as with the previous test I added, the idea is to add a call to the subroutine (in this case `atrcpt`) directly in a test program after `setim8b` has been called. I think the main issue here is that I don't know what outcome I'm looking for. Should I be looking for certain values in my modified bufr message, or should I be content that this runs without anything crashing?

test/outtest4.F90
```diff
@@ -134,4 +134,10 @@ program outtest4
   ! Close the output file.
   call closbf ( 13 )
 
+  ! Test atrcpt, which should add 6 bytes to mgbf
+  ilena = iupbs01(mgbf, 'LENM')
+  call atrcpt(mgbf, lmgbf, mgbf)
+  ilenb = iupbs01(mgbf, 'LENM')
+  IF ((ilenb-ilena) .ne. 6) stop 3
+
```
jbathegit, line 143: There's already a "stop 3" branch further up in the code.  This one should be "stop 6".

test/outtest4.F90
```diff
@@ -6,7 +6,7 @@
 program outtest4
   implicit none
 
-  integer*4 isetprm, ireadsb, igetmxby, icbfms
+  integer*4 isetprm, ireadsb, igetmxby, icbfms, ilena, ilenb, iupbs01
```
jbathegit, line 10: FYI, only iupbs01 itself needs to be declared as integer\*4.

ilena and ilenb can just be declared as integer, because it's OK to assign an integer\*4 return value from iupbs01 into either one of them, no matter whether they happen to be integer\*4 or integer\*8 themselves.

test/outtest4.F90
jbathegit: I also just noticed that you're passing in mgbf as both the 1st and 3rd arguments.  These really should be separate arrays as noted in the docblock for atrcpt; in fact, I'm surprised you didn't end up with some sort of memory overwrite error here.

So probably best to go back to a separate array for the 3rd argument, like the separate mgbfapp array you had earlier, and just so this doesn't bite us unexpectedly at some point down the road.

test/outtest4.F90
```diff
@@ -134,4 +134,10 @@ program outtest4
   ! Close the output file.
   call closbf ( 13 )
 
+  ! Test atrcpt, which should add 6 bytes to mgbf
+  ilena = iupbs01(mgbf, 'LENM')
+  call atrcpt(mgbf, lmgbf, mgbf)
+  ilenb = iupbs01(mgbf, 'LENM')
+  IF ((ilenb-ilena) .ne. 6) stop 3
```
edwardhartnett: Extra parens not needed.

jbathegit:
Hi Alex, did you see any of my earlier comments from last Friday?  It doesn't look like you've made any such updates.

For example, this new test should have a "stop 6" instead of "stop 3", and you shouldn't pass the same array as the 1st (input) and 3rd (output) arguments to atrcpt().

test/outtest4.F90
jbathegit: Or, if #414 ends up getting merged first, then this update can become "stop 7" :-)

AlexanderRichert-NOAA:
> Hi Alex, did you see any of my earlier comments from last Friday? It doesn't look like you've made any such updates.
> 
> For example, this new test should have a "stop 6" instead of "stop 3", and you shouldn't pass the same array as the 1st (input) and 3rd (output) arguments to atrcpt().

I don't see any comments from you on this PR if that's what you mean. Any chance you wrote a review but didn't complete it?

I fixed the two items you mentioned.

jbathegit:
Hmm, that's weird, I've commented on a lot of PRs in the past, but I've never had to explicitly click "Send" before!?  I must have gotten into some strange window when I did those comments, b/c looking back I see they all still had a Pending label on them!?

Anyway, I've just re-posted, and hopefully you can see them now.  Sorry about that!

jbathegit:
The only comment I just added now was the one about "stop 7" vs. "stop 6".  The others I had originally tried to send on Friday.

jbathegit (approved):

PR #412, opened 2023-03-24
modernize C-Fortran interface for crwbmg routines

jbathegit:
Part of #79
Part of #252
Part of #254


edwardhartnett:
WRT doxygen I ran into this problem before and it's a weird one. Here's a stackoverflow question that gives the answer:
https://stackoverflow.com/questions/59970141/how-to-get-doxygen-to-work-with-fortran-interface-statements

So I updated the code to include that, and we'll see what the CI has to say...

jbathegit:
Wow, thanks Ed - I'm not sure I would have ever found that on my own!

As you can see, the CI ran fine, and this also brings the gcovr up to exactly 80.0%, which is a good milestone!

and there was much rejoicing!

The [rendered Doxygen output](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/index.html) also looks pretty-good too, so if you can approve this PR I'll go ahead and get it merged.  And I'll go ahead and get #410 merged as well.

edwardhartnett (approved):

PR #414, opened 2023-03-25
Add igetdate 8-byte test to outtest4.F90

AlexanderRichert-NOAA:
PR completes testing for `igetdate`. Checked date values with debufr.


test/outtest4.F90
```diff
@@ -131,6 +132,10 @@ program outtest4
 
   call writsa ( -13, mxbfmg, mgbf, lmgbf )
 
+  ! Get Section 1 date (returns 8-byte ints for KIND_8).
+  call igetdate(mgbf, mear, mmon, mday, mour)
+  if (mear.ne.20 .or. mmon.ne.10 .or. mday.ne.1 .or. mour.ne.11) stop 6
```
edwardhartnett: Nice! I like the way you checked all values, not just some of them.

edwardhartnett (dismissed):

test/outtest4.F90
jbathegit: Thanks Alex, but just FYI that igetdate is a function, not a subroutine.  So it has a return value that could also be checked, but at a minimum it should at least be called as a function (not as a subroutine), and it should also be explicitly declared as "integer*4 igetdate" at the top of the routine, alongside of isetprm, ireadsb, igetmxby, and icbfms.

jbathegit:
I just merged #411, which means there's now a merge conflict with this PR b/c they both have changes to outtest4.  But I think you should be able to clear these up pretty quickly, and just make this igetdate a "stop 7" instead of "stop 6", since atrcpt is now using the latter.

AlexanderRichert-NOAA:
@jbathegit yep just merged

jbathegit (approved):

PR #415, opened 2023-03-25
more bort tests, some documentation fixes, some misc tests

edwardhartnett:
Part of #381 
Part of #397 

Testing more borts().

### DIFF
--- a/src/readerme.f
+++ b/src/readerme.f
@@ -12,42 +12,37 @@ C> might otherwise be used, and from that point on, the application
 C> program can proceed with a call to one of the
 C> [subset-reading subroutines](@ref hierarchy) (and then,
 C> subsequently, to any of the
-C> [values-reading subroutines](@ref hierarchy)), just
-C> like if readmg() had been called instead.
+C> [values-reading subroutines](@ref hierarchy)).
 C>
-C> When using this subroutine, it's still necessary for the
+C> When using this subroutine, it's necessary for the
 C> application program to have previously called subroutine openbf() in
 C> order to associate a DX BUFR tables file with the message that is
-C> being input via MESG, and it's still also necessary to pass in the
+C> being input via MESG; it's also necessary to pass in the
 C> relevant LUNIT value as a call argument, even though in this case
 C> the subroutine will not actually try to read from the associated
 C> Fortran logical unit.
 C>
 C> If MESG contains a DX BUFR table message, the subroutine will
 C> store the contents internally and use them to process any
-C> future BUFR messages associated with LUNIT.  In this case, the
+C> future BUFR messages associated with LUNIT. In this case, the
 C> subroutine will return with IRET = 11, and any number of
 C> DX BUFR table messages passed in via consecutive calls to this
 C> subroutine will accumulate internally and be treated as a single DX
 C> BUFR table, up until a call is made where MESG no longer contains a
 C> DX BUFR table message.
 C>
-C> @param[in] MESG    -- integer(*): BUFR message
-C> @param[in] LUNIT   -- integer: Fortran logical unit number for
-C>                       BUFR file
-C> @param[out] SUBSET  -- character*8: Table A mnemonic for type of BUFR
-C>                        message that was read
-C>                        (see [DX BUFR Tables](@ref dfbftab)
-C>                        for further information about Table A mnemonics)
-C> @param[out] JDATE   -- integer: Date-time stored within Section 1 of
-C>                        BUFR message that was read, in format of either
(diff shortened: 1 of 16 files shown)


src/readerme.f
```diff
@@ -12,42 +14,37 @@
...
+C> @param[out] SUBSET - character*8: Table A mnemonic for type of BUFR
+C> message that was read (see [DX BUFR Tables](@ref dfbftab)
+C> for further information about Table A mnemonics).
+C> @param[out] JDATE - integer: Date-time stored within Section 1 of
+C> BUFR message that was read, in format of either YYMMDDHH or YYYYMMDDHH,
+C> depending on the most recent call to datelen().
+C> @param[out] IRET - integer: return code:
+C> - 0 MESG was successfully read.
+C> - 11 MESG contained a DX BUFR table message.
+C> - -1 MESG contained an unrecognized Table A message type.
```
jbathegit, line 47: Once again for the record, IMO the removal of all of this whitespace makes the docblock a lot less readable.  Not everyone relies on just the rendered html from Doxygen for documentation - some of us still occasionaly edit the actual library routines and read it there.

src/uptdd.f
```diff
@@ -3,16 +3,21 @@
 C>
 C> @author J. Woollen @date 1994-01-06
 
-C> This subroutine returns the bit-wise representation of the FXY value corresponding to, sequentially,
-C> a particular (IENT'th) "child" mnemonic of a Table D sequence ("parent") mnemonic.
+C> Returns the integer FXY value corresponding to a child
+C> mnemonic of a Table D sequence parent mnemonic.
```
jbathegit, line 7: We've always taken care to explicitly call this the "bit-wise representation of the FXY value", rather than just an "integer FXY value". 

For example, if the character representation of the descriptor is "063022", then one could wrongly assume that the corresponding "integer FXY value" is just 63022, but instead what we're really talking about here is a value of 16150.  See the remarks in the docblock for ifxy() for a detailed example.

Bottom line, there's a reason we've always referred to the latter value as the "bit-wise representation of the FXY value" throughout the entire library, rather than just  the "integer FXY value" :-)

jbathegit:
BTW, I noticed in the developer CI run that a number of your new ufbstp tests were failing due to a SIGSEGV, rather than failing within the bort routine like we want:

`Program received signal SIGSEGV: Segmentation fault - invalid memory reference.`

Also, at least two new tests are failing with memory leaks:

```
1: ==6968==ERROR: LeakSanitizer: detected memory leaks
1: 
1: Direct leak of 32 byte(s) in 1 object(s) allocated from:
1:     #0 0x7fb0dc424867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
1:     #1 0x7fb0dc0b7ef8  (/lib/x86_64-linux-gnu/libgfortran.so.5+0x22ef8)
1: 
1: Indirect leak of 512 byte(s) in 1 object(s) allocated from:
1:     #0 0x7fb0dc424867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
1:     #1 0x7fb0dc0b7ef8  (/lib/x86_64-linux-gnu/libgfortran.so.5+0x22ef8)
1: 
1: SUMMARY: AddressSanitizer: 544 byte(s) leaked in 2 allocation(s).
```

Of course it can be really hard to notice these things with test_bort, since the roles are reversed and a fail now gets interpreted as a pass.  So I think we really need to get in the habit of digging through the developer run output to check for these.

src/uptdd.f
edwardhartnett: OK, the problem is that "bit-wise representation of the FXY value" is quite confusing. Even with your explanation, I don't understand the different between 63022 and 16150. Let's discuss at the meeting today...

edwardhartnett:
OK, let's separate the bort testing from the documentation changes, in order to divide and conquer this PR. ;-)

I have put the tests (after fixing the ufbstp() seg faults) in new PR #422 .

Once those are merged I will circle around again to this PR. We will have a documentation meeting today to discuss some of the issues with the documentation changes in this PR.

jbathegit:
Thanks Ed, but I'm still seeing the aforementioned memory leaks in the test output for the developer CI run.  If you pop open the details for the test step and search for `ERROR: LeakSanitizer`, you'll see that there are a couple of them, right between tests for upftbv case 2 and wrdxtb case 1.

edwardhartnett:
The testing work in the PR has been submitted in #422 

We will discuss some documentation issues at today's meeting, and then I will resumbit some of the documentation work as future PRs...

PR #416, opened 2023-03-25
adding test_c_misc.c

edwardhartnett:
In this PR I add a test_c_misc.c for the same purpose as the fortran test_misc.F90, and add one line of test code that covers an uncovered line in the rbytes function.


test/test_c_misc.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This file tests misc C API stuff.
+ *
+ * Ed Hartnett 3/25/23 */
+
+#include <stdio.h>
+#include "bufr_interface.h"
+
+int rbytes(char *bmg, int mxmb, int isloc, int newbytes);
+
```
jbathegit, line 11: Just FYI, the reason there wasn't any existing prototype for function rbytes is b/c it's not intended to ever be called directly from an application program.  In fact, it's only ever meant to be called from inside of function crbmg, so I just defined it inside of the same source file (i.e. crwbmg.c) as function cread, and therefore I didn't even need to create a prototype for it in the first place ;-)

That said, I agree if you're going to try calling it directly from inside of this new test_c_misc.c, then you're going to need to define a prototype for it as you've done here.

jbathegit:
Thanks Ed - this provides a nice infrastructure for further testing of outlier cases inside of the C part of the library!

jbathegit (approved):

PR #417, opened 2023-03-25
added module doc

edwardhartnett:
Part of #397 

@jbathegit I forgot to mention this in code review, but we need doxygen documentation for every module as well. Even if it's a repeat of the file documentation block, as in this case.

Doxygen needs a brief description of the module as well as the file.

### DIFF
--- a/src/bufr_interface.F90
+++ b/src/bufr_interface.F90
@@ -4,6 +4,10 @@
 !>
 !> @author J. Ator @date 2023-03-22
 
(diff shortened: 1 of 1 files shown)


jbathegit:
OK, but I don't think the module description should have a @brief tag.  That tag should only be in the file documentation block, correct?

How about I try to add a more detailed module description as part of my next PR, rather than just repeating a copy of the same brief description from the file documentation block?  Then you could just close this PR.

PR #422, opened 2023-03-28
More bort tests, some misc tests

edwardhartnett:
Part of #381

These are the new bort tests from PR #415.

Also some misc tests.

@jbathegit I have fixed the segfaults in ufbstp() bort testing. That's for catching them.

This contains the tests from #415 without the documentation changes. Let's get the tests merged and doing some good while we continue to discuss documentation issues.

### DIFF
--- a/test/run_test_bort.sh
+++ b/test/run_test_bort.sh
@@ -37,22 +37,39 @@ for kind in "4" "d"; do
 
     # Check copybf().
     (./test_bort_$kind copybf 1) && exit 1
+    (./test_bort_$kind copybf 2) && exit 1
 
     # Check copymg().
     (./test_bort_$kind copymg 1) && exit 1
+    (./test_bort_$kind copymg 2) && exit 1
+    (./test_bort_$kind copymg 3) && exit 1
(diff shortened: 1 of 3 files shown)


jbathegit:
Thanks Ed, but I'm still seeing the aforementioned memory leaks in the test output for the developer CI run. If you pop open the details for the test step and search for `ERROR: LeakSanitizer`, you'll see that there are a couple of them, right between tests for upftbv case 2 and wrdxtb case 1.

edwardhartnett:
OK, stand by while I take a look. But I would expect memory leaks from test_bort, because instead of closing anything, it just aborts. ;-)

edwardhartnett:
OK, I have fixed the memory leaks.

In general, that may not always be possible, and that's OK. test_bort can leak memory, and the memory leak will cause a non-zero error code, but the script run_test_bort.sh is already expecting a non-zero return code, so it does not cause the CI testing to fail.

If memory leaks occur tests other than test_bort.F90, then they will cause CI to fail. So this is all was we wish it to be.

jbathegit (approved):

PR #425, opened 2023-03-29
finish upgrading utility tests

### LINKED ISSUE #376: Simplify util tests

jbathegit:
Fixes #376

Part of #254

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -109,7 +109,7 @@ list(APPEND test_c_interface_srcs
   test_c_interface_2.c
 )
 
-# Here's a C test.
+# test_c_misc test.
 set(test_name test_c_misc)
 add_executable(${test_name} ${test_name}.c)
 add_dependencies(${test_name} bufr_4)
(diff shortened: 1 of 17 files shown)


edwardhartnett (approved):
Nice simplification of CMake for utils testing.

PR #426, opened 2023-03-29
more documentation edits

edwardhartnett:
Part of #397


src/readsb.f
```diff
@@ -3,31 +3,28 @@
...
-C> current message, and therefore that a new call needs to be made to
-C> one of the [message-reading subroutines](@ref hierarchy) in order
-C> to read in the next message from logical unit LUNIT.
+C> Whenever this subroutine returns with IRET = 0, this indicates that a
+C> new BUFR data subset (i.e. report) was successfully read into internal
+C> arrays within the BUFRLIB software, and from where it can be
+C> manipulated or further parsed via calls to any of the [values-reading
+C> subroutines](@ref hierarchy).  If the subroutine returns with IRET =
+C> -1 there are no more data subsets available within the current
+C> message; a new call needs to be made to one of the [message-reading
```
jbathegit: Be careful here - I've found that if you lead off a line with a dash, then Doxygen interprets it as the start of a bulleted list.  So probably best to just add the "-1" to the end of the previous comment line.

jbathegit: Otherwise I think this all looks OK.

edwardhartnett: OK, I've reworked that a bit.

jbathegit (approved):

Issue #230, opened 2022-08-26
update Python documentation via pdoc


jbathegit:
@aerorahul @jswhit @JackReevesEyre @edwardhartnett do any of you have access to the pdoc utility or know a way to run it?  I'd like to update the Python documentation for this library, because we now have a new read_long_string() wrapper courtesy of Jack's work on #226, and it already includes the proper Markdown syntax for pdoc.  But I don't know how to run pdoc myself, because I can't find it anywhere on my local machine nor on WCOSS2, even when I have the latest Python module loaded.

For some background, our existing Python documentation is available from https://noaa-emc.github.io/NCEPLIBS-bufr/python/index.html, but that was provided to me in the past (I believe by Rahul, though I don't recall for sure), rather than being generated by anything that I know how to run myself.  Or maybe I'm just looking for pdoc in the wrong place and need a quick primer on where/how to run this myself(?)  

I'd appreciate any help or suggestions - thanks!

JackReevesEyre:
Didi you manage to sort this @jbathegit ? I am not familiar with the pdoc package. I see that the pdoc web page (https://pdoc.dev/) only mentions pip install, but it also seems to be available on conda-forge (https://anaconda.org/conda-forge/pdoc) if that's preferable. Then running `pdoc /<rest of path>/NCEPLIBS-bufr/python/ncepbufr/` might (?) then create the new html pages equivalent to https://noaa-emc.github.io/NCEPLIBS-bufr/python/.

jbathegit:
Thanks for the reply @JackReevesEyre.  I did some more reading, and at least on WCOSS2 I made some headway finding pdoc:

```
jeff.ator@clogin05 [~ (158)] $ module load python/3.8.6
jeff.ator@clogin05 [~ (159)] $ pip install pdoc
Defaulting to user installation because normal site-packages is not writeable
Collecting pdoc
  Downloading pdoc-12.1.0-py3-none-any.whl (127 kB)
     ââââââââââââââââââââââââââââââââââââââââ 127.2/127.2 kB 898.9 kB/s eta 0:00:00
Collecting MarkupSafe
  Downloading MarkupSafe-2.1.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (25 kB)
Collecting pygments>=2.12.0
  Downloading Pygments-2.13.0-py3-none-any.whl (1.1 MB)
[...]

PR #428, opened 2023-03-30
Clean up user_guide and other .md files

edwardhartnett:
Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/397

This weirdly fixes the problems of lots of additional entries in the modules list. Turns out in our .md files we need the first title to use "#" and the next level to use "##". If instead we use "##" and "###", then doxygen gets confused.

Also used emacs auto-indent to clean up line-lengths and formatting.


docs/user_guide.md
```diff
@@ -1,17 +1,109 @@
...
+  <td>ufbint() ufbrep() ufbseq() ufbstp() ufbint_f() ufbrep_f() writlc() setvalnb() setbmiss()</td>
+</tr>
+</table>
 
-The Python API documentation is available
-[here](https://noaa-emc.github.io/NCEPLIBS-bufr/python/index.html); however,
-note that this functionality is only available if the library was built using
-the -DENABLE_PYTHON=ON option during the cmake step.
+Many of the above subroutines and functions can also be accessed via a separate
+[Python API](https://noaa-emc.github.io/NCEPLIBS-bufr/python/index.html).
```
jbathegit: This seems redundant - do we still need this here since you've already got a link to the same Python API at the top of this file?

In any case, I think the note about "this functionality is only available if the library was built using the -DENABLE_PYTHON=ON option during the cmake step." is useful and should probably be retained.

jbathegit: In other words, I think wherever we reference the Python API it's important to note that this functionality is only available if that particular option was switched on during the build.

jbathegit:
BTW, thanks for all of this!  I had actually noticed all of those additional entries in the modules list, and I had been planning to fix it as part of my next PR, although using two '##' characters everywhere instead of one '#' for the title in each .md file and '##' for the rest of the places.  But this seems fine.

Just let me know about the above Python question, and once that's resolved I'll go ahead and approve and merge.

docs/user_guide.md
edwardhartnett: Instead of saying that in the C/Fortran documentation, let's say that in the python documentation.

Honestly, no one will be surprised that the python package is optionally built. That's quite common.

I have removed the reference to the python docs at the bottom of the page.

jbathegit:
OK, but I'll just point out that right now we don't have any way to update the Python (pdoc) documentation.  We have an issue #230 that's been open for this for a long time.

docs/user_guide.md
jbathegit: OK, but I'll just point out that right now we don't have any way to update the Python (pdoc) documentation. We have an issue https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/230 that's been open for this for a long time.

jbathegit (approved):

PR #429, opened 2023-03-30
more documentation

edwardhartnett:
Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/397

Better brief descriptions for some subprograms.


src/wtstat.f
```diff
@@ -1,9 +1,10 @@
 C> @file
-C> @brief Update the status of a system file with respect to the BUFRLIB software.
+C> @brief Open a BUFR file, or update file status.
 C>
```
jbathegit, line 3: I think this is misleading, because it doesn't really "Open a BUFR file".  I think the brief and the below comment should both just be "Update a file status with respect to the library", or just "Update a file status" if you think the first one is too  long.  But again "Open a BUFR file" is misleading, because this routine only updates a file status w.r.t. the library - it doesn't actually "Open" it.

src/writdx.f
```diff
@@ -1,21 +1,25 @@
...
-C> @param[in]  lun   -- integer: i/o stream index into internal memory arrays
-C> @param[in]  lundx -- integer: fortran logical unit number containing
-C>                               dictionary table information to be used (by readdx()) to
-C>                               create internal tables written to lunit;
-C>                               if set equal to lunit, this subroutine calls bort()
+C> @param[in] lunit - integer: fortran logical unit number for BUFR
+C> file being written.
+C> @param[in] lun - integer: i/o stream index into internal memory
+C> arrays.
+C> @param[in] lundx - integer: fortran logical unit number containing
```
jbathegit, line 18: Should we go ahead and change this to "File ID" like we discussed yesterday, rather than "i/o stream index into internal memory arrays"?

src/wtstat.f
edwardhartnett: OK, but this function says:

C> This subroutine can be used to connect or disconnect a specified
C> Fortran logical unit number to/from the BUFRLIB software, and it
C> can also be used to set or reset the internal message status
C> associated with that logical unit number.

Doesn't that mean it does an open like openbf()?

src/writdx.f
edwardhartnett: OK.

src/wtstat.f
jbathegit: No, it doesn't, and I can see how the existing description may be misleading.  This routine only updates the status of one of the internal I/O indices (now "file ID") based on the output from a recent call to status(), which in turn gets called from within openbf().

But bottom line, no this routine doesn't actually open a file like openbf(), or fortran_open(), or cobfl().

jbathegit: In other words, in this case "connect or disconnect" means it officially registers the presence of this logical unit within its internal file ID array.  But again, there's no "open" or "close" ever going on in this routine.

edwardhartnett: But if the file has already been opened by openbf(), then isn't the file already in the internal file array? In other words, doesn't it already have a file ID?

jbathegit: I understand it can be a bit confusing, but openbf internally calls status (to get a new file ID) and then wtstat (to actually record the new file ID).  But then wtstat also gets called later by other routines when, e.g. a new message is opened or closed for that file ID.

So all wtstat ever does is just update internal information (i.e. status information ;-) corresponding to a file ID.  It never actually opens a file.

edwardhartnett: OK, got it, let me see if I can say it better...

edwardhartnett: Maybe in the brief description of the function we should also indicate whether it is internal? In other words, not intended for the user to call directly?

edwardhartnett: OK, I have rephrased. Take a look.

jbathegit: Correct, wtstat is only intended to be used internally and should never be called directly by a user.

jbathegit (approved):

PR #434, opened 2023-03-31
Add testing for writsa when compression enabled

### LINKED ISSUE #359: complete testing for writsa.f

AlexanderRichert-NOAA:
This PR adds testing for `writsa` when `CCMF=='Y'`.

Fixes #359

### DIFF
--- a/test/outtest9.F90
+++ b/test/outtest9.F90
@@ -36,7 +36,9 @@ program outtest9
  
   implicit none
 
-  integer ii, jj, imgdt, ier
+  integer ii, jj, imgdt, ier, mxbfmg, lmgbf
+  parameter ( mxbfmg = 50000 )
+  integer mgbf ( mxbfmg )
 
   integer*4 nmsub, icopysb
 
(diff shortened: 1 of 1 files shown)


test/outtest9.F90
```diff
@@ -101,7 +103,11 @@ program outtest9
     if ( ier .ne. 2 ) stop 9
 
     ! Write the subset into the output message.
-    call writsb (21)
+    if ( ii.lt.3 ) then
+      call writsb (21)
+    else
+      call writsa (21, mxbfmg, mgbf, lmgbf)
```
edwardhartnett, line 109: Are we checking this result somehow? Does it alter the output in any way?

AlexanderRichert-NOAA: On that last pass of the loop, I'm using `writsa` as a replacement for `writsb`, which shouldn't and doesn't change the results. So the checking is the existing logic down below as far as number of messages and all that.

jbathegit: Good work Alex - I like how you figured out how to test this using an existing test code!

jbathegit (approved):

PR #435, opened 2023-03-31
Add testing for ufbcnt in 8-byte mode

### LINKED ISSUE #366: complete testing of ufbcnt.f
The I8 testing is missing.

AlexanderRichert-NOAA:
Fixes #366

### DIFF
--- a/test/intest7.F90
+++ b/test/intest7.F90
@@ -38,7 +38,7 @@ program intest7
 
   integer*4 isetprm, igetprm, ireadns, ibfms
 
-  integer imgdt, iret, jdate, nr8v, idx, nsub, num
+  integer imgdt, iret, jdate, nr8v, idx, nsub, num, kmsg, ksub
 
   integer mxr8pm, mxr8lv
   parameter ( mxr8pm = 15 )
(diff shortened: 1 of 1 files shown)


test/intest7.F90
```diff
@@ -136,10 +136,13 @@ program intest7
   call ufbseq ( 11, r8val, 0, 1, nr8v, 'CLON' )
   idx = index( errstr(1:errstr_len), 'UFBSEQ - 3rd ARG. (INPUT) IS .LE. 0' )
   if ( ( nr8v .ne. 0 ) .or. ( idx .eq. 0 ) ) stop 17
+  ! Test ufbcnt for 8-byte mode.
+  call ufbcnt(11, kmsg, ksub)
```
jbathegit, line 140: FWIW, this comment is a little misleading since the new ufbcnt test will end up running for all modes, not just 8-byte. 

It's fine to run it for all modes, but I assume you realize that, if we really only wanted it to run for 8-byte, then we'd need to encase it in an `#ifdef KIND_8` wrapper ;-)

jbathegit (approved):

test/intest7.F90
AlexanderRichert-NOAA: Yes, I was being lazy about my description :) If there's a better way to phrase it I'm happy to change it.

PR #404, opened 2023-03-20
remove the mysterious -w flag

edwardhartnett:
When I remove the -w flag, we get a bunch of Fortran warnings.

PR #437, opened 2023-04-03
more cleanup for eventual removal of -w flag

jbathegit:
Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/300

Part of https://github.com/NOAA-EMC/NCEPLIBS-bufr/pull/404


src/nemspecs.f
```diff
@@ -120,7 +120,7 @@ RECURSIVE SUBROUTINE NEMSPECS
      .          ( TAGN(1:8) .EQ. TAGNRV(JJ) ) .AND.
      .          ( NODE .GE. ISNRV(JJ) ) .AND.
      .          ( NODE .LE. IENRV(JJ) ) ) THEN
-              NREF = NRV(JJ)
+              NREF = INT(NRV(JJ))
```
edwardhartnett, line 123: Sometimes int sometimes nint? 

When not specified, int is used, right? So are you sure all the nint() calls are correct?

edwardhartnett (approved):

src/nemspecs.f
jbathegit: NINT is for converting from a real to an integer.  INT is used when converting from an integer\*8 to an integer\*4.

PR #444, opened 2023-04-11
added tests 

edwardhartnett:
Part of #445

Added some more test code to cover some uncovered cases...


test/intest1.F90
```diff
@@ -34,6 +35,9 @@ program intest1
   call setim8b(.true.)
 #endif
 
+  ! Test the i4dy() function.
+  if (i4dy(80123023) .ne. 1980123023) stop 1
+
```
jbathegit: There's already a `stop 1` line further down below in the code, so could you please renumber this to a different stop code, or else leave this and your below `if (idxmsg(1) .ne. 0) stop 2` lines as is and instead just renumber all of the existing stop numbers 1-18 as 3-20?  Or just move these two new tests to the end of the code and number them as 19 and 20?

Either way is fine with me, but the idea is to have a unique number for every stop error, so we always know immediately which test failed, correct?

test/test_misc.F90
```diff
@@ -46,6 +50,20 @@ program test_misc
...
+  imsg = 1
+  if (ireadmm(imsg, subset, idate) .ne. 0) stop 4
+  if (imsg .ne. 2 .or. subset .ne. 'ADPSFC' .or. idate .ne. 23022519) stop 5
+  call closbf(11)
+
+  ! Test iupb().
+  mbay(1) = 1
+  mbay(2) = 2
+  if (iupb(mbay, 1, 1) .ne. 0) stop 6
+
```
jbathegit, line 68: Similar to my above comment for intest1.F90, this test_misc.F90 already has `stop 3`, `stop 4`, and `stop 5` lines in the code, and the idea is for stop numbers to be unique in a test code, correct?

edwardhartnett: OK

test/intest1.F90
edwardhartnett: OK

test/intest1.F90
```diff
@@ -25,6 +25,7 @@ program intest1
        bfmg(mxbf), cds3(mxds3)*6, tagpr*8, celem*60, cunit*22
   character*20 filnam / 'testfiles/IN_1' /
   character filost / 'r' /
+  integer i4dy, idxmsg
```
jbathegit: The i4dy and idxmsg functions should be explicitly declared as integer*4, similar to the other library functions in line 16 which return integer values.

Remember the only library build now is with 4-byte integers, so all functions which return integer values will always return them as 4-byte integers.

test/test_misc.F90
```diff
@@ -9,7 +9,11 @@ program test_misc
   integer lun, il, im
   integer ios
   integer num, iret
-
+  integer ireadmm, imsg, idate, iunit
+  integer mbay(2)
```
jbathegit, line 14: Similar to the above comment, the ireadmm function should be explicitly declared as integer*4 here, since the below ireadmm test isn't encased in an `#ifdef KIND_4` block and therefore runs for all of the _4, _8, and _d tests.

Contrast this with the igetprm function declaration in line 24, which is encased in an `#ifdef KIND_4` block and therefore only ever runs for the _4 test, and so for that reason it's OK to just declare that one as integer.

test/intest1.F90
edwardhartnett: OK, done.

test/test_misc.F90
edwardhartnett: OK, done.

jbathegit (approved):

PR #446, opened 2023-04-11
more bort tests

edwardhartnett:
Part of #381 

There's more bort testing to be done, but these tests are good and ready to be merged...


jbathegit:
FYI, case 2 of ufbqcp is faulting with a SIGSEGV

edwardhartnett:
OK, thanks for catching that. There's only one bort() in ufbqcp() so there should not even be a second test case. I've removed it.

.github/workflows/developer.yml
```diff
@@ -52,6 +52,9 @@ jobs:
     - name: test
       run: |
         cd bufr/build
+        cd test
+        bash -x run_test_bort.sh
+        cd ..
         ctest --verbose --output-on-failure --rerun-failed
```
jbathegit: Why is run_test_bort.sh now being run directly in the developer.yml script?  There's already an add_test directive in CMakeLists.txt which runs the run_test_bort.sh script for all of the builds, so we now end up running it twice in the developer CI build.  Why is this necessary?

test/CMakeLists.txt
```diff
@@ -147,6 +149,16 @@ foreach(kind ${test_kinds})
   endforeach()
 endforeach()
 
+# Create some C tests.
+create_c_test_plain(test_c_misc)
+
+# The C bort test is not rul as a regular test, because it fails.
+set(test_name test_c_bort)
```
jbathegit, line 169: rul -> run (typo in comment)

.github/workflows/developer.yml
edwardhartnett: I will fix. this was an artifact of debugging...

test/CMakeLists.txt
edwardhartnett: Fixeed.

jbathegit (approved):

PR #447, opened 2023-04-12
Add cnved4 testing to outtest2.F90

### LINKED ISSUE #348: complete testing of cnved4.f
This code is untested:

```
C       Check for I8 integers.

        IF(IM8B) THEN
           IM8B=.FALSE.

           CALL X84 ( LMSGOT, MY_LMSGOT, 1 )
           CALL CNVED4 ( MSGIN, MY_LMSGOT*2, MSGOT )

           IM8B=.TRUE.
           RETURN
        ENDIF

        IF(IUPBS01(MSGIN,'BEN').EQ.4) THEN

C         The input message is already encoded using edition 4, so just
C         copy it from MSGIN to MSGOT and then return.

          NMW = NMWRD(MSGIN)
          IF(NMW.GT.LMSGOT) GOTO 900
          DO I = 1, NMW
            MSGOT(I) = MSGIN(I)
          ENDDO
          RETURN
        ENDIF

```

AlexanderRichert-NOAA:
Completes testing for cnved4.

Fixes #348


test/outtest2.F90
```diff
@@ -102,7 +104,14 @@ program outtest2
   ! Confirm the "missing" value is still the same value that was set previously via the call to setxmiss.
   IF ( nint(xmiss) .ne. nint(getbmiss()) ) stop 3
 
-  call writsb ( 11 )
+  ! Test cnved4 to cover im8b=.true. case
+  call writsa ( 11, mxbfmg, mgbf, lmgbf )
+  mgbf2 = mgbf
+  call cnved4(mgbf,mxbfmg,mgbf2)
```
jbathegit, line 109: Why do you need to copy mgbf into mgbf2 in line 109?  The call to cnved4 in line 110 will already do that itself, won't it?

test/outtest2.F90
```diff
@@ -102,7 +104,14 @@ program outtest2
...
+  if ( iupbs01(mgbf2, 'BEN') .ne. 4 ) stop 4
+  ! Re-converting to BUFR ed. 4 should leave the message unchanged
+  call cnved4(mgbf2,mxbfmg,mgbf)
+  if ( .not. all( mgbf(1:lmgbf) .eq. mgbf2(1:lmgbf) )) stop 5
```
jbathegit, line 113: I'm also not sure I understand why you're calling cnved4 a second time in line 113?  Unless I'm missing something, it seems to me you should be able to just remove line 113, and I think the test in line 114 should still work.

jbathegit: Or is the message that comes out of writsa in the mgbf array initially encoded as edition 3?

AlexanderRichert-NOAA: Correct, it's initially ed 3, so the second pass is needed to test the logic of "converting" from ed 4 to ed 4.

test/outtest2.F90
AlexanderRichert-NOAA: Just removed the `mgbf2 = mgbf`

jbathegit (approved):

Issue #362, opened 2023-03-11
complete testing of ufbrw.f


edwardhartnett:

jack-woollen:
@edwardhartnett  I made a branch called jsw_ufbrw_test as a place to start. I put the code in the utils directory because that's the simplest way to include in the cmake process. It probably belongs somewhere else but i'd like some guidance on that.  There's a script in test_scripts to run the test. It checks many of the possible operations in the ufbrw subroutine, both input and output.

edwardhartnett:
Awesome! Can we remove the script and just have the test program do everything?

jack-woollen:
Done. Might now be ready for prime time.

edwardhartnett:
We have some decent testing for this subroutine (65% coverage) but there are some large untested error conditions. Can we reach these in a test? @jack-woollen what do you think?

jack-woollen:
I think so. I'll work on it.

PR #448, opened 2023-04-14
Jsw ufbrw test

jack-woollen:
I made a branch called jsw_ufbrw_test as a place to start. I put the code in the utils directory because that's the simplest way to include in the cmake process. It probably belongs somewhere else but i'd like some guidance on that. There's a script in test_scripts to run the test. It checks many of the possible operations in the ufbrw subroutine, both input and output.
This is meant to address issue #362.


test/test_scripts/test_ufbrw_test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
edwardhartnett: Do we need this script? Can the program just be run from CMake like the other tests?

jack-woollen: Yup, looks like it can. Working on it...

jack-woollen: Done and working. Just need a little cleanup and a doc block for it.

test/test_ufbrw.F90
```diff
@@ -0,0 +1,180 @@
+! This is a test for NCEPLIBS-bufr library subroutine ufbrw and its callees.
+!
+! This tests reads and writes a testfile exercising the mechanics in subroutine
+! ufbrw which locate and extract or install values from or into the bufr
+! testfile and test various user defined options available through subroutine
+! ufbint which is the driver of ufbrw.
+!
+! J Woollen  4/14/23
```
edwardhartnett, line 8: Thanks for the excellent documentation.

test/test_ufbrw.F90
```diff
@@ -0,0 +1,180 @@
+! This is a test for NCEPLIBS-bufr library subroutine ufbrw and its callees.
+!
+! This tests reads and writes a testfile exercising the mechanics in subroutine
+! ufbrw which locate and extract or install values from or into the bufr
+! testfile and test various user defined options available through subroutine
+! ufbint which is the driver of ufbrw.
+!
+! J Woollen  4/14/23  
+
+      program tstwin
```
edwardhartnett: Program name incorrect here.

test/test_ufbrw.F90
```diff
@@ -0,0 +1,180 @@
...
+      data brr(48) /"998.00 ******** ********     4.60     2.20"/
+      data brr(49) /"850.00 ******** ********     2.00    -1.70"/
+      data brr(50) /"700.00 ******** ********    12.10    -4.40"/
+      data brr(51) /"998.00  8112.00    22.30 ******** ********"/
+      data brr(52) /"925.00  6312.00    14.10 ******** ********"/
+      data brr(53) /"850.00  2161.00    14.80 ******** ********"/
+      data brr(54) /"700.00  2131.00     9.20 ******** ********"/
+      data brr(55) /"300.00 ******** ******** ******** ********"/
+
+      print *, 'Testing misc subroutines.'
```
edwardhartnett: Edit this print statement to something like:

Testing ufbrw() subroutine.

test/test_ufbrw.F90
```diff
@@ -0,0 +1,180 @@
...
+
+      file="testfiles/data/prepbufr2.ref"
+
+! open the print file to record ufbrep test results
+
+      open(55,file='ufbrw_prnt_out')
+
+! test various reading options applying user specified filtering, testing conwin
+
+      do i=1,5
```
edwardhartnett, line 90: Does your editor not do autoindenting? I would expect indenting inside do loops...

test/test_ufbrw.F90
```diff
@@ -0,0 +1,180 @@
...
+      call strsuc(brr(n),str2,len2)
+      if(n<=55.and.iret==0.and.str1/=str2) then
+         print*,n,len1,len2
+         print*,"str1:",str1  
+         print*,"str2:",str2 
+         call bort('error exit from ufbrw_test')
+      elseif(n>55.and.iret==0) then
+         print*,n,iret
+         call bort('error exit from ufbrw_test')
+      elseif(n==56) then
```
edwardhartnett: What's this elsif for? Won't the do loop already end without it when n is 56?

jack-woollen: checking to see if the copy made to many lines.

test/test_ufbrw.F90
```diff
@@ -0,0 +1,179 @@
...
+          print*,"str2:",str2 
+          call bort('error exit from ufbrw_test')
+        elseif(n>55.and.iret==0) then
+          print*,n,iret
+          call bort('error exit from ufbrw_test')
+        elseif(n==56) then
+          exit
+        endif
+      enddo
+
```
jbathegit, line 170: Thanks for this Jack.

The only comment I would add is that we've taken a different approach in all of our other test codes, which is to Fortran "stop" immediately with a unique non-zero number whenever there's an unexpected result, (and which in turn causes the test to automatically fail), rather than continuing to run through to the end of the code.

So I would recommend changing each of your `call bort('error exit from ufbrw_test')` lines to just be `stop n` and that way you're guaranteed a unique non-zero number which tells you immediately which line of brr didn't match its expected counterpart.  And then just change your `exit` line to `stop 57`.

And that way you also don't ever need to print the value of n either, because the stop number will already tell you that.

test/test_ufbrw.F90
jack-woollen: Thanks for the good advice.

test/test_ufbrw.F90
jack-woollen: I prefer straight edges but did the indents anyway.

test/test_ufbrw.F90
jack-woollen: That works.

jbathegit (approved):

Issue #445, opened 2023-04-11
add tests to increase test code coverage


edwardhartnett:
add tests to increase test code coverage

This is a catch-all issue for test improvements that are not part of a specific test issue.

edwardhartnett:
We are making great progress in testing.

As of today, we have 84.5% test coverage, up almost 10% from when we started this test campaign. Thanks to @AlexanderRichert-NOAA , @jbathegit , and @jack-woollen for the great effort here.

We still have a ways to go. When I look at the code coverage report, I still see a lot of red...

![image](https://user-images.githubusercontent.com/38856240/234861706-a7788ac8-5cd7-45a0-98cb-e8ec0942d1a1.png)

![image](https://user-images.githubusercontent.com/38856240/234861863-a93c6598-d01c-4787-857f-551fbb34a629.png)

edwardhartnett:
Tremendous progress here:

![image](https://github.com/NOAA-EMC/NCEPLIBS-bufr/assets/38856240/644978bb-30f9-424e-8ae7-0af05763a69f)

jbathegit:
Testing is now at 95.5%, so we should probably consider closing this issue

![Capture6](https://github.com/NOAA-EMC/NCEPLIBS-bufr/assets/62722702/1383f04d-b688-4f2e-9abf-b580e85827a3)

edwardhartnett:
Awesome!

![image](https://github.com/NOAA-EMC/NCEPLIBS-bufr/assets/38856240/11a366cd-3b65-4765-aad8-69a6de482f74)

PR #449, opened 2023-04-15
testing cread.c functions

edwardhartnett:
Part of #445 

Add tests for cread.c code, including testing failures.


edwardhartnett:
This PR increases testing for cread.c functions, but it does more.

It also adds simple tests for cread.c. These tests should be run before more complex system-level tests. Any bug introduced into cread.c will be debugged much more quickly from test_c_cread.c than from any system level testing. In the system level tests, the programmer has to start by locating the bug. With test_c_cread.c that's already done, and that is often the most time-consuming part of debugging.

test/test_c_cread.c
```diff
@@ -0,0 +1,183 @@
...
+
+    /* Allocate storage to read a BUFR message into. Without the +8
+     * these buffers will be overflowed by the cdrbufr(). See
+     * https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/450. */
+    if (!(msg = malloc(MSG_SIZE + 8)))
+        return TEST_ERR;
+    if (!(msg2 = malloc(MSG_SIZE + 8)))
+        return TEST_ERR;
+        
+    /* Open a real BUFR file, read a BUFR message, and close the file. */
```
jbathegit, line 55: As I noted in the discussion of #450, I really don't think there's a need to do this.  I think the problem here is that you're confusing integers and bytes.  The msg and msg2 arrays are correctly declared as integer arrays, but the 3rd argument to crdbufr (i.e. mxwrd) is the number of **integer** elements of the array, not the number of bytes.  So instead of passing in MSG_SIZE as the 3rd argument to crdbufr, you should really be passing in MSGSIZE/4 (or MSGSIZE/sizeof(int), as below).

jbathegit: If you do that (without allocating the 8 extra bytes) and it still fails, can you let me know what the return code from crdbufr is?  If it's -3 then I think what I'll do is tweak the logic to only check for nintrem+3>mxwrd if nbytrem%nb>0.  In other words, if the BUFR message size in bytes is an exact multiple of 4 (as in the case of 14928), then we only really need to check for nintrem+2>mxwrd as the overflow case, and maybe that's what's really going on here.

Let me know, but again in any case I don't think you should ever have to purposely allocate more space than you actually need to fit the largest size BUFR message you want to read.

edwardhartnett: I tried your change of checking nbytrem%nb. This allows the function to proceed, but address sanitizer still tells me I am overflowing here:

```
    bufr[nintrem + 1] = wkint[0];
    bufr[nintrem + 2] = wkint[1];

```
Not sure why.

jbathegit: As noted, I'll take a closer look at this in #450

jbathegit (approved):

Issue #349, opened 2023-03-11
complete testing of copysb.f


edwardhartnett:
Some of this untested code is error handling, but there is some untested code about handling an oversized subset that can be tested.

edwardhartnett:
This is now tested to 84% coverage and only some error conditions are not tested.

PR #456, opened 2023-04-22
added more CI testing

jbathegit:
Part of #445
Part of #349
Part of #33

This increases total test coverage to 83.7% (lines) and 66.1% (branches), by filling in some missing testing gaps in copysb, cpyupd, iok2cpy, ipkm, lmsg, mesgbc, msgupd, nmwrd, and ufbcpy.

The i8bm (8-byte integer wrapper) is now also fully tested for ipkm, lmsg, mesgbc, and nmwrd.


jack-woollen (approved):

edwardhartnett (approved):

src/status.f
```diff
@@ -81,8 +72,17 @@ RECURSIVE SUBROUTINE STATUS(LUNIT,LUN,IL,IM)
 C  SEE IF UNIT IS ALREADY CONNECTED TO BUFR ARCHIVE LIBRARY SOFTWARE
 C  -----------------------------------------------------------------
 
+      IF ( .NOT. ALLOCATED(IOLUN) ) THEN
```
edwardhartnett, line 75: I think this should call bort().

I don't see how the user can proceed with any function which tries to call status() before initialization of the internal arrays. So even if status() issues a warning and returns, there will still be a segfault when the function calling status tries to use internal memory.

jack-woollen: Since status returns lun=0 in this case, the program will know that openbf has not been called yet and could then do that. In other words it could be asking that question. But there is ambiguity because lun returned as zero could also mean there is no more table space to open a file, which is a different matter with a different solution. Lun could return -1 in one case to clear that up. Plus there should also be at least an errwrt if there's no more file table space. Thanks @edwardhartnett for pointing out the contradiction. And certainly borting is also a reasonable alternative for either problem.

PR #458, opened 2023-04-24
change to test CI

edwardhartnett:


src/cread.c
```diff
@@ -28,7 +28,7 @@
  */
 void
 openrb(int nfile, char *ufile) {
-   pb[nfile] = fopen( ufile, "rb " );
+   pb[nfile] = fopen(ufile, "rb ");
 }
```
jbathegit, line 32: This was already "fixed" in #449, so why are you doing this here?

PR #465, opened 2023-04-26
gets1loc testing

### LINKED ISSUE #352: complete testing of gets1loc.f
This function sez:

```
C> This subroutine returns the location of a specified value within
C> Section 1 of a BUFR message.

```

Looks like reading a test file and making sure the correct answers are returned would be useful.

The I8 code is not tested here.

Also several  error cases are not tested. These can be tested in the normal way (i.e. run the function and look at return value to ensure failure). Since bort is not used in this function we don;t have to worry about it here.

AlexanderRichert-NOAA:
This PR completes test coverage for gets1loc.
- Test `YCEN` and `CENT` labels
- Test attempting nonexistent value
- Test `IM8B=.true.` case

Fixes #352

### DIFF
--- a/test/test_misc.F90
+++ b/test/test_misc.F90
@@ -14,6 +14,7 @@ program test_misc
   integer mbay(2)
   character*8 subset
   integer iupb
+  integer isbyt, iwid
   
 #ifdef KIND_4
   character*5 char5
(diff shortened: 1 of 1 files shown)


test/test_misc.F90
```diff
@@ -65,6 +66,19 @@ program test_misc
   mbay(1) = 1
   mbay(2) = 2
   if (iupb(mbay, 1, 1) .ne. 0) stop 11
+
+  ! Test gets1loc for YCEN and CENT positions
+  call gets1loc('YCEN', 4, isbyt, iwid, iret)
+  if ( iret .ne. -1 ) stop 12
+  call gets1loc('CENT', 4, isbyt, iwid, iret)
```
edwardhartnett, line 73: IN all these cases we should be checking the values of all output parameters, not just iret. There's also:

```
C> @param[out] ISBYT   -- integer: Number of starting byte within Section 1
C>                        which contains value corresponding to S1MNEM
C> @param[out] IWID    -- integer: Width (in bits) of value corresponding
C>                        to S1MNEM, counting from the first bit of the
C>                        byte pointed to by ISBYT

```

Always test all the output parameters for every test.

edwardhartnett (approved):

PR #467, opened 2023-04-27
adding test for stseq()

edwardhartnett:
Part of #445 

adding test for stseq()

### DIFF
--- a/test/test_c_stseq.c
+++ b/test/test_c_stseq.c
@@ -7,6 +7,12 @@
 #include <stdio.h>
 #include "bufr_interface.h"
 #include "bufrlib.h"
+#include "cread.h"
+
+#define IN_9 "testfiles/IN_9"
+#define TEST_ERR 1
+#define MSG_SIZE 14928 /* in bytes */
+#define MSG_SIZE_INT ( MSG_SIZE / sizeof(int) ) /* in ints */
 
 /* Prototype. */
 void stseq(f77int *lun, f77int *irepct, f77int *idn, char *nemo,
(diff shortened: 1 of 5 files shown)


edwardhartnett:
OK, I am trying to call the stseq() function. I open the file with openrb(), and then I try to call stseq(). Here's my test code:

```
    /* Open a real BUFR file, read a BUFR message, and close the file. */
    openrb(0, IN_9);
    if (!pb[0])
        return TEST_ERR;
    if (crdbufr(0, msg, MSG_SIZE_INT))
        return TEST_ERR;

    stseq(&lun, &irepct, &idn, nemo, cseq, cdesc, &ncdesc);

```

Within stseq() the Fortran subroutine NUMTBD(). It segfaults on this code:

 `        DO I=1,NTBB(LUN)
`

because array NTBB is not allocated.

What function should I be calling which will cause NTBB to get allocated? @jbathegit @jack-woollen any suggestions?

jack-woollen:
@edwardhartnett   call openbf(0.'QUIET',0) will alloc the arrays.

test/test_c_stseq.c
```diff
@@ -0,0 +1,60 @@
...
+
+#define IN_9 "testfiles/IN_9"
+#define TEST_ERR 1
+#define MSG_SIZE 14928 /* in bytes */
+#define MSG_SIZE_INT ( MSG_SIZE / sizeof(int) ) /* in ints */
+
+/* Prototype. */
+void stseq(f77int *lun, f77int *irepct, f77int *idn, char *nemo,
+           char *cseq, f77int *cdesc, f77int *ncdesc);
+
```
jbathegit, line 20: A prototype for this function is already in bufrlib.h.  You shouldn't need to also list it here.

PR #469, opened 2023-04-27
more testing for msgupd, hold4wlc, and pkbs1

jbathegit:
Part of #445

### DIFF
--- a/src/arallocf.f
+++ b/src/arallocf.f
@@ -37,6 +37,7 @@ C> @author J. Ator @date 2014-12-04
         USE MODV_MXS01V
         USE MODV_MXTAMC
         USE MODV_MXTCO
+        USE MODV_MXH4WLC
         USE MODV_NFILES
 
         USE MODA_USRINT
@@ -72,6 +73,7 @@ C> @author J. Ator @date 2014-12-04
         USE MODA_BITMAPS
         USE MODA_NRV203
         USE MODA_RLCCMN
+        USE MODA_H4WLC
 
         CHARACTER*80 ERRSTR
         CHARACTER*36  BRTSTR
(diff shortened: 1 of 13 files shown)


AlexanderRichert-NOAA (approved):

jack-woollen (approved):

src/arallocf.f
```diff
@@ -562,5 +566,16 @@ SUBROUTINE ARALLOCF
...
 
+C       MODA_H4WLC arrays.
+
+        ALLOCATE( LUH4WLC(MXH4WLC), STAT=iost )
+        IF ( iost .ne. 0 ) CALL BORT( BRTSTR // 'LUH4WLC' )
+
+        ALLOCATE( STH4WLC(MXH4WLC), STAT=iost )
+        IF ( iost .ne. 0 ) CALL BORT( BRTSTR // 'STH4WLC' )
+
+        ALLOCATE( CHH4WLC(MXH4WLC), STAT=iost )
```
edwardhartnett, line 577: This is not just testing, right?

Did you also find and fix some bugs? Or refactor?

jbathegit: I decided to include MXH4WLC as one of the settings that can be configured by the user via isetprm, rather than fixed at some arbitrary parameter value as before.  This did make it easier to test the remaining statements in hold4wlc, but I also think it's a useful overall upgrade to the code, and which is why I went ahead with it.

So this wasn't to fix any sort of bug.  Instead, it was just an upgrade to make the library a bit more flexible/configurable for users.

edwardhartnett: It would be good in such a case to add a new issue which describes  the refactor/improvement.

PR #476, opened 2023-05-03
more testing for cpdxmm()

### LINKED ISSUE #33: Unit tests for src/copybf.f, src/copymg.f, src/copysb.f, src/cpbfdx.f, src/cpdxmm.f, src/cpymem.f, src/cpyupd.f

edwardhartnett:
Part of #445 
Fixes #33 

More testing for cpdxmm().

With this test, all codes mentioned in #33 have pretty good testing. A giant improvement from our starting point, where all of these functions were completely untested!

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -58,7 +58,7 @@ jobs:
         mkdir build
         cd build
         cmake -DTEST_FILE_DIR=/home/runner/data -DCMAKE_INSTALL_PREFIX=./install -DENABLE_PYTHON=ON ..
-        make -j2 VERBOSE=1
+        make VERBOSE=1
         make install
         cd python
         sudo python3 setup.py install
--- a/test/test_misc.F90
+++ b/test/test_misc.F90
@@ -104,7 +104,13 @@ program test_misc
   call strnum('    ',num,iret)
   if ((iret .ne. 0) .or. (num .ne. 0)) stop 403
 
-
+  ! Testing cpdxmm().
+  open(unit = 11, file = 'testfiles/IN_2', form = 'UNFORMATTED', iostat = ios)
+  if (ios .ne. 0) stop 5
+  call openbf(11, 'IN', 11)
+  call openbf(11, 'QUIET', 2)
+  call cpdxmm(11)
+  call closbf(11)
 
   ! These tests only for the _4 run of test_misc.
 #ifdef KIND_4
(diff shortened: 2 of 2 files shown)


edwardhartnett:
@jbathegit and @jack-woollen this code seems to work fine, but how do I tell:

```
  ! Testing cpdxmm().
  open(unit = 11, file = 'testfiles/IN_2', form = 'UNFORMATTED', iostat = ios)
  if (ios .ne. 0) stop 5
  call openbf(11, 'IN', 11)
  call cpdxmm(11)
  call closbf(11)

```

After cpdxmm() has been called, what function(s) can I call to confirm that the embedded DX table has been correctly copied into memory?

jack-woollen (dismissed):
In the test you commented out, the way its written, unit 12 needs to point to an ASCII version of the bufr tables. The file IN_1 is a bufr file with no bufr tables of any kind in it.

edwardhartnett:
@jack-woollen thanks, but that does not answer my question. ;-)

How do I check that cpdxmm() has correctly copied to memory? What function can I call to learn the contents of that memory?

jack-woollen:
My comment was about a different part of the code. Now that you mention cpdxmm I say two things. That test is not setup correctly because there's no memory file open. If you open a memory file (call ufbmem...) then its fairly complicated to check the result of calling cpdxmm. It would take some trial and error for me to work it out. I'm surprised cpdxmm doesn't check for an open memory file. Maybe Jeff should weigh in. @jbathegit

jbathegit:
We could certainly add a failsafe check to cpdxmm, but for the record I'll point out that cpdxmm isn't a routine that was ever intended to be called directly by an application code.  In fact, the only place I know of that it's currently called from is within ufbmem itself, which by definition means that a file has already been opened to the internal memory arrays, and therefore such a failsafe check isn't really needed anyway.

So I'm struggling now to recall why we even included an im8b wrapper inside of cpdxmm in the first place.  I presume that's what Ed is trying to test in this PR, correct?  Jack, do you know of any application codes which call cpdxmm directly, and therefore why we even need an im8b wrapper inside of that routine?  If not, then I'd vote to just take it out of there.

jack-woollen:
I think Ed's issue was about how to check to see if a test worked. Whether it needs to be tested or not is another issue I guess.

edwardhartnett:
@jbathegit indeed it is the handling of the im8b wrapper that I am testing here.

I will add an issue for it's removal...

Issue #257, opened 2023-01-10
remove the _8 build


edwardhartnett:
The idea of doing different builds for different sized default integers or floats is a bad idea.

Instead, we should use modern Fortran features to handle different types of data. As every other Fortran project does.

We intend to get rid of _4 and _d builds eventually, but we can get rid of the _8 build now. No one uses it.

jbathegit:
This is no longer an issue as of PR #250 which was merged last month.  There's now only one build, and this is now reflected in the README.md as well.

edwardhartnett:
OK, but the idea of removing the  _8 build was not to make code changes and add a bunch of code to support it. It was to remove it and have it be gone completely.

Users can still build the _8 style library (along with their own codes) by setting the compiler flags themselves.

The intention was that _8 would be removed from the build system. Not that it would be added to the code.

jbathegit:
@edwardhartnett that was not at all how @jack-woollen and I understood the issue.  In all of our prior discussions, the plan was to modify the existing code so that only one build of the library would be needed going forward which could seamlessly work with _8 and _d application codes.  Updating the Fortran code to be able to remove all of those extra builds had always been the goal as we understood it.  And the _8 and _d have indeed been removed from the build system - there's no reference to them anywhere in any of the CMake files, except for the test subdirectory so we can verify that the one remaining build continues to work for any application codes built with _8 and _d compiler flags.

If what you've now described was really your vision all along, then it would have been extremely helpful if you could have chimed in with that back in February of last year when we were having these initial discussions in #78, and even more so during June-August of last year when we were having more detailed discussions about how to resolve this in #217 and were asking/begging for feedback from you and others on the various approaches we had come up with.  But we heard nothing but crickets, so we went ahead and agreed on a way forward between the two of us.  Are you really jumping back in now to tell us that all of the work we did to resolve this was a complete waste of our time?

jack-woollen:
[...]

PR #488, opened 2023-05-04
more C-Fortran interface updates

jbathegit:
Part of #79

### DIFF
--- a/src/restd.c
+++ b/src/restd.c
@@ -35,11 +35,9 @@
 */
 
 void
-restd(f77int *lun, f77int *tddesc, f77int *nctddesc, f77int *ctddesc)
+restd(int *lun, int *tddesc, int *nctddesc, int *ctddesc)
 {
-    f77int i0 = 0;
-
-    f77int desc, ncdesc, cdesc[MAXNC];
+    int desc, ncdesc, cdesc[MAXNC];
     int i, j, inum, itbd, ictbd;
     int iscl, iref, ibit;
 
@@ -49,15 +47,15 @@ restd(f77int *lun, f77int *tddesc, f77int *nctddesc, f77int *ctddesc)
 **  How many child descriptors does *tddesc have?
 */
     numtbd_f(*lun, *tddesc, nemo, NEMO_STR_LEN+1, &tab, &itbd);
-    uptdd( &itbd, lun, &i0, &inum );
+    uptdd_f(itbd, *lun, 0, &inum);
 
     *nctddesc = 0;
 /*
 **  Examine each child descriptor one at a time.
 */
     for ( i = 1; i <= inum; i++ ) {
-        uptdd( &itbd, lun, &i, &desc );
-        if (! istdesc( &desc ) ) {
+        uptdd_f(itbd, *lun, i, &desc);
+        if (! istdesc_f(desc) ) {
 /*
 **          desc is a local descriptor.
 */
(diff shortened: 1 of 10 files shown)


jack-woollen (approved):

src/wrdesc.c
```diff
@@ -15,16 +15,16 @@
...
+ * written into descary.
+ * @param descary - Array of descriptors.
+ * @param ndescary - Number of descriptors written so far into
  * descary.
  *
  * @author J. Ator @date 2004-08-18
  */
 void
-wrdesc(f77int desc, f77int *descary, f77int *ndescary)
+wrdesc(int desc, int *descary, int *ndescary)
```
edwardhartnett, line 27: How come we are changing these? Is there an issue to describe this change?

src/uptdd.f
```diff
@@ -10,18 +10,17 @@
 C> For a description of the WMO bit-wise representation of the FXY
 C> value, see ifxy().
 C>
-C> @param[in] ID - integer: positional index of parent mnemonic
+C> @param[in] ID - integer: Positional index of parent mnemonic
```
edwardhartnett, line 13: Is there another kind of index than positional? I suggest you simple say "index of parent mnemonic..."

src/stseq.c
```diff
@@ -125,14 +125,14 @@ stseq(f77int *lun, f77int *irepct, f77int *idn, char *nemo,
 **  private copy is created and dynamically allocated during each
 **  recursive call to this subroutine.
 */
-    f77int *rpdesc;
+    int *rpdesc;
 
 /*
 **  The following variables are declared as static so that they
```
edwardhartnett, line 131: In C it's not really normal to have your comments indented differently from the rest of the code. Normally comments are indented to the same level as the code they apply to.

src/restd.c
```diff
@@ -35,11 +35,9 @@
 */
```
edwardhartnett, line 35: What I really don't get is why some parts of the library are in C, and other parts in Fortran. How is the decision made as to what language to use for what?

In other libraries, C forms the lowest level of code, and is called by Fortran, but I've never seen the opposite anywhere else except NCEPLIBS-bufr. That is, I don't know of any other libraries where C code calls Fortran code.

src/wrdesc.c
jbathegit: This f77int typedef is a remnant from back when we had multiple _4, _8, etc. builds of the library, as a way to easily link a C "int" to a Fortran integer (in the case of the _4 build) vs. linking a "long" to a Fortran integer (in the case of the _8 build).  That way, we could use a single typedef throughout the C source code that would work for either case.  But now, following the resolution of #78 and #257, there's only a single _4 library build now, and so f77int is always an "int" and therefore the typedef is superfluous and can go away.

So, in a nutshell, this is basically remaining cleanup from the consolidation down to a single _4 library build, but on the C side which we're still trying to get cleaned up and thereby simplify the code and make it cleaner and easier to understand.

And to answer your earlier question about why we need to change these C function signatures in the first place, that goes back to #79 where we're trying to, among other things, "remove the UNDERSCORE and all related shenanigans" (your words ;-).  Modernizing this C-Fortran interface using BIND also means that we no longer need to pass extra string length arguments when calling Fortran from C, which is yet another change to many of these function signatures.  And while doing so I'm also following the earlier model that Ron M. laid out where you append _f when calling Fortran from C, and _c when calling C from Fortran.  So just trying to keep everything consistent throughout the library.

src/restd.c
jbathegit: For better or worse, this is just how the library evolved over many years/decades.  In some cases (e.g. the logic in restd.c and stseq.c) we needed to use recursion, which Fortran77 didn't universally support on all platforms at the time.  Similarly, not all platforms supported the reading and writing of binary BUFR data in Fortran without the use of kludgy extensions or blocking mechanisms, and older versions of Fortran didn't always support things like structures, or intrinsic functions such as qsort and bsearch which were needed for sorting and searching through large BUFR master tables.  But since all of this functionality was readily available in standard C, and since there was a universally-supportable way to link Fortran to C (albeit using the aforementioned UNDERSCORE and other "shenanigans" ;-), then the decision was made to use C where and when it was needed in the library.

I get that more modern versions of Fortran (95, 2003), etc. can now pretty-much do everything that C can do, and as you know we've been taking advantages of these modern features as we continue to maintain the library, but at this point I'm not keen to go back and rewrite all of the existing C functionality in Fortran just for consistency with other libraries.

edwardhartnett: OK, that makes sense.

src/wrdesc.c
edwardhartnett: OK, makes sense.

PR #491, opened 2023-05-05
removed im8b section from cpdxmm()

### LINKED ISSUE #483: remove im8b handling from subroutine cpdxmm()?
In cpdxmm() we have:

```
C       Check for I8 integers

        IF(IM8B) THEN
           IM8B=.FALSE.

           CALL X84(LUNIT,MY_LUNIT,1)
           CALL CPDXMM(MY_LUNIT)

           IM8B=.TRUE.
           RETURN
        ENDIF

```

This code is currently untested (see #476 ).

Should this code be there? That is, should cpdxmm() even handle IM8B?

If it should be removed, I will remove it and abandon the testing added in #476. If not, I will get the test in #476 working better.

edwardhartnett:
Fixes #483 

@jack-woollen and @jbathegit make sure this is what we want to do before merging this PR.

### DIFF
--- a/src/cpdxmm.f
+++ b/src/cpdxmm.f
@@ -10,12 +10,11 @@ C> @param[in] LUNIT   -- integer: Fortran logical unit number for BUFR
 C>                       file
 C>
 C> @author J. Ator @date 2009-03-23
-        RECURSIVE SUBROUTINE CPDXMM( LUNIT )
+        SUBROUTINE CPDXMM( LUNIT )
 
         use bufrlib
 
         USE MODV_MXDXTS
-        USE MODV_IM8B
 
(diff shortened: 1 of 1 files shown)


jbathegit:
I think it is.  Jack, can you think of any reason why cpdxmm would ever need to be called directly from an application code, and therefore why it would need to have an im8b wrapper?  I can't, as I noted previously in #476.

BTW Ed, and for sake of simplicity, you can also remove the "RECURSIVE" qualifier from the SUBROUTINE statement, and you can also remove the "USE MODV_IM8B" statement, as neither of those would be required any longer either.

jbathegit (approved):

PR #496, opened 2023-05-09
more testing and cleanup

### LINKED ISSUE #485: docblock errors in ufbint, ufbget, ufbevn, and ufbrep?
@edwardhartnett's recent updates to test_ufbcup.F90 seem to have uncovered some errors in our documentation.

Specifically, I don't see any code logic in the library that allows the following to be true in ufbint.f, ufbget.f, ufbevn.f, or ufbrep.f:

```
C> - If ABS(LUNIN) points to a file that is open for input (reading
C> BUFR), there are a few additional special mnemonics that can be
C> included within STR when calling this subroutine, and which in turn
C> will result in special information being returned within the
C> corresponding location in USR. These special mnemonics are not
C> considered to be part of Table B or Table D and therefore do not
C> need to be definied within the DX or master table file associated
C> with ABS(LUNIN):
C>      - NUL - returns the "missing" value
C>      - IREC - returns the number of the BUFR message within the
C>               file pointed to by ABS(LUNIN) (counting from the
C>               beginning of the file) in which the current data
C>               subset resides
C>      - ISUB - returns the number of the current data subset within
C>               the BUFR message pointed to by IREC, counting from
C>               the beginning of the message
```

In fact, I've looked at our past releases going all the way back to v10.2.5 in 2019, and I don't see any such functionality anywhere in those versions either which would allow the above statements to be true.  In fact, the only routines where I see that this could work are ufbtab.f and ufbtam.f, though for those it's only for IREC and ISUB (and another variable ITBL for ufbtam.f), but never for NUL.

@jack-woollen am I missing something here?  If not, then my suggestion is that we correct the docblocks for ufbint.f, ufbget.f, ufbevn.f, and ufbrep.f so they no longer mislead users into thinking these are valid STR mnemonics for those routines.

jbathegit:
Fixes #485

Part of #445, including for pktdd, rdmemm, rdusdx, and seqsdx.

### DIFF
--- a/src/rdusdx.f
+++ b/src/rdusdx.f
@@ -3,6 +3,8 @@ C> @brief Read a complete DX BUFR table.
 C>
 C> @author Woollen @date 1994-01-06
 
+C> Read a complete DX BUFR table.
+C>
 C> This subroutine reads and parses a file containing a user-
 c> supplied DX BUFR table in character format, and then stores
 c> this information into internal arrays in module @ref moda_tababd.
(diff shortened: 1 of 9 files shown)


edwardhartnett (approved):
Nice. Love the bort tests!

jack-woollen (approved):
Good call on the n>ntag test and error message in seqsdx. Looks right now.

PR #498, opened 2023-05-12
more modernization of C-Fortran interface

### LINKED ISSUE #79: Use modern fortran/C API

### LINKED ISSUE #266: can we remove code dealing with EBCDIC?
I think EBCDIC is dead.

Are there any platforms that don't use ASCII?

Can we remove all code dealing with EBCDIC?

jbathegit:
Fixes #79
With this PR, this long-running issue is now complete!

Fixes #266
As I was looking to change the C-Fortran interface for ichkstr, and since we're no longer worrying about EBCDIC, I realized I could just use strncmp to do the same thing as ichkstr in C, and in turn I was able to completely eliminate the need for both the ichkstr and chrtrna routines within the library.  From there, it was simple enough to go ahead and remove the remaining lines of code in the library which support EBCDIC.


jbathegit:
@jack-woollen @edwardhartnett @AlexanderRichert-NOAA would one of you mind looking this over so I can get it merged? It's been almost a week, and I have more commits in the pipeline to improve the testing coverage, but I'd really prefer to do those as a separate PR.

jack-woollen (approved):

PR #501, opened 2023-05-23
more testing for iupm, stndrd, rdmems, nenubd, parstr, and parusr

jbathegit:
Part of #445


src/ipkm.f
```diff
@@ -3,15 +3,19 @@
...
-C> @param[in] NBYT   -- integer: Number of bytes of CBAY (up to a
-C>                      maximum of 8) within which to encode N
-C> @param[out] CBAY  -- character*(*): String of length NBYT bytes
-C>                      containing encoded integer N
+C> This subroutine is the logical inverse of subroutine iupm().
+C>
+C> @param[in] N - integer: Value to be encoded.
+C> @param[in] NBYT - integer: Number of bytes of CBAY (up to a
+C> maximum of 8) within which to encode N.
+C> @param[out] CBAY- character*(*): String of length NBYT bytes
```
AlexanderRichert-NOAA, line 17: Not sure if the lack of space before the hyphen will mess up the doxygen parsing

AlexanderRichert-NOAA (approved):

src/ipkm.f
jbathegit: Good catch.  I'll make sure I get that fixed in my next PR.

PR #502, opened 2023-05-24
more testing for mtfnam

jbathegit:
Part of #445

### DIFF
--- a/src/mtinfo.f
+++ b/src/mtinfo.f
@@ -43,7 +43,6 @@ C> @author J. Ator @date 2009-03-23
 
       CHARACTER*(*) CMTDIR
 
-      CHARACTER*128 BORT_STR
       CHARACTER*240 MTDIR
 
 C-----------------------------------------------------------------------
--- a/src/mtinfo.f
+++ b/src/mtinfo.f
@@ -63,7 +63,6 @@ C  ---------------------
       ENDIF
 
       CALL STRSUC ( CMTDIR, MTDIR, LMTD )
-      IF ( LMTD .LT. 0 ) GOTO 900
 
       LUN1 = LUNMT1
       LUN2 = LUNMT2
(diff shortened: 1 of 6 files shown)


src/mtinfo.f
```diff
@@ -72,6 +70,4 @@ RECURSIVE SUBROUTINE MTINFO ( CMTDIR, LUNMT1, LUNMT2 )
 C  -----
 
       RETURN
-900   BORT_STR = 'BUFRLIB: MTINFO - BAD INPUT MASTER TABLE DIRECTORY:'
```
AlexanderRichert-NOAA, line 75: Is removing these lines related to testing or something else?

jbathegit: I was trying to test for this in test_bort.  But I quickly realized that, with the changes to strsuc in #406, it's no longer possible for that routine to return with LMTD<0, which in turn means that lines 75 and 76 can no longer be reached within this routine.

I could have changed the if test to LMTD<=0, but I realized that whatever values get returned for MTDIR and LMTD will always be quickly checked anyway (via an inquire statement inside of subroutine mtfnam), so this check in strsuc is now basically superfluous and can be removed.

AlexanderRichert-NOAA (approved):

jack-woollen:
Jeff

I just got this from cloning/compiling a bufrlib branch on wc2.

/lfs/h2/emc/global/noscrub/Jack.Woollen/cris-fix/NCEPLIBS-bufr/test/test_misc.F90(207):
error #6345: Either a PAUSE, STOP or ERROR STOP statement has an invalid
argument.
    if ( (iret .ne. 0) .or. (igetprm(trim(prms(i))) .ne. 42+i) ) stop 800+i

It doesn't like stop 800+i !

Jack

AlexanderRichert-NOAA:
Ah dang it, that was me. I forgot some compilers/fortran versions don't like that. I'm looking at this file right now so I'll fix it while I'm at it.

jbathegit:
Hey Jack, just curious - what compiler were you trying to use on WC2?  If it was the Intel compiler, then it's surprising that this hasn't already cropped up during the GitHub CI runner testing.

In short, I agree we need to get this fixed, but just wondering why it never showed up in the CI test for Intel?  More of a curiosity as much as anything.

PR #504, opened 2023-05-26
Add test coverage; fix test_misc 'stop' issue

AlexanderRichert-NOAA:


jbathegit:
This looks fine, and thanks for this @AlexanderRichert-NOAA , but could you or @jack-woollen please remind me what compiler has a problem with a non-constant argument to the "stop" command?  Is it the Intel compiler on WCOSS2, and if so I wonder why this has never shown up with the Intel test under the GitHub CI runner?

jbathegit (approved):

jack-woollen:
This is the fortran version on wcoss2 that doesn't like variable stop values.  intel/19.1.3.304

PR #509, opened 2023-06-02
exit gracefully from bad replication factor in usrtpl

### LINKED ISSUE #495: BUFRLIB: RDMSGW - SKIPPING CORRUPTED MESSAGE
@ilianagenkova @jbathegit @jack-woollen  @ am seeing this error in the satellite ingest logs. Please check here:
4;m[clogin07 /lfs/h2/emc/obsproc/noscrub/sudhir.nadiga/CRIS/ingest_logs]$ ls -l tranjb_ingest_cris_fsr_431.out
-rw-r--r-- 1 sudhir.nadiga obsproc 9930707 May  8 03:42 tranjb_ingest_cris_fsr_431.out

Running go_chkdat on the tank for that day also shows this error. This may be related on problems that the GSI
is having reading the files. Andrew is reporting this error:
/scratch1/NCEPDEV/stmp2/Andrew.Collard/ROTDIRS/v16.3.5_n21_cris/logs/2023040418/gdasanal.log.4
The error is on line 14998:
 ***********BUFR ARCHIVE LIBRARY ABORT**************
 BUFRLIB: USRTPL - REPLICATION FACTOR OVERFLOW (EXCEEDS MAXIMUM OF   255 ({GCRCHN}  )
***********BUFR ARCHIVE LIBRARY ABORT**************

We will begin investigating this error now, and we hope to find a sturdy solution. Any help from Jack and Jeff is welcome. 

Jack had emailed a temporary solution. 

-------------------------- JACK WOOLLEN TEMPORARY SOLUTION --------------------------------------
After puzzling over the corrupted message warnings, it turns out they are being generated when tranjb is reading through the tank file in order to append new data onto the end. In other words, there is pre-existing garbage in the tank files, probably left over from failed runs during testing. I notice a number of tranjb aborts among the logs. Garbage in a tank can be removed easily! The script example below shows how. If you run it twice you will see the garbage has disappeared.

Jack

set -euax
clean=/lfs/h2/emc/global/noscrub/Jack.Woollen/jwfork/build/utils/split_by_subset
tank=/lfs/h2/emc/obsproc/noscrub/sudhir.nadiga/SCRIPTDIR/SEMIOPS/TEMPNEW/20220803/b021/xx223
rm -f NC021223; $clean $tank; mv NC021223 $tank

----------------------------------------------------------------------------------------------------------------

We can first try to see if this solves the problem for Andrew and if we can incorporate it in the satingest code. 
Any alternate solutions (e.g., increasing the default value for the problematic bufrlib variable) can also be
investigated.

jbathegit:
Fixes #495


src/usrtpl.f
```diff
@@ -186,10 +187,12 @@ SUBROUTINE USRTPL(LUN,INVN,NBMP)
       CALL BORT(BORT_STR)
 904   WRITE(BORT_STR,'("BUFRLIB: USRTPL - REPLICATION FACTOR OVERFLOW'//
      . ' (EXCEEDS MAXIMUM OF",I6," (",A,")")') JVAL,TAG(NODI)
-      CALL BORT(BORT_STR)
+      call errwrt(bort_str)
```
AlexanderRichert-NOAA, line 190: Is there a short/simple answer to why this returns a code as opposed calling bort as in the other cases? Is it to do with the context of the failure in operations (i.e. the calling programming needs to keep running)?

jbathegit: Yes, as noted in #495, this particular error stopped the operational GSI code dead in its tracks, so we needed to trap and move past that error without aborting the program.

As noted in https://github.com/NOAA-EMC/NCEPLIBS-bufr/discussions/340, we have a longer-term goal to refactor the library so that it never aborts an application program.  But that will be a longer-term task, so for now this PR can be viewed as a prototype for one way we might go about doing that.

AlexanderRichert-NOAA (approved):

PR #515, opened 2023-07-19
Update python/CMakeLists.txt to fix Python version

climbfuji:
The current logic in `python/CMakeLists.txt` detects incorrect Python versions. `3.10` gets truncated to `3.1`.

This fixes it


jbathegit (approved):

climbfuji:
Thanks @jbathegit for merging this, and if you could please notify us when you have a new release to add to spack, that would be greatly appreciated.

PR #523, opened 2023-09-13
final updates for new release 12.0.1

### LINKED ISSUE #516: wrong name for tarball
I don't know how it happened but the 12.0.0 tarball is named: NCEPLIBS-bufr-bufr_v12.0.0.tar.gz 

Let's make sure the next release has the correct style of name: NCEPLIBS-bufr-12.0.0.tar.gz is what I expected.

I don;t believe we should change the 12.0.0 tarball, but make sure this is right for the next release.

@AlexanderRichert-NOAA for the current bufr release in spack we have:

    url = "https://github.com/NOAA-EMC/NCEPLIBS-bufr/archive/refs/tags/bufr_v11.5.0.tar.gz"

So this needs to be updated and also the new incorrect tarball name needs to be accomodated. I'm not sure how to do that, and perhaps we should fix the existing tarball name - what do you think Alex?

### LINKED ISSUE #521: Need new release tag for c readlc extension.
Unfortunately we will need a new release tag so that the readlc extension can be made available via the various software stacks. It could just be version 12.0.1 or something as the change is very minor.

Please see: https://github.com/NOAA-EMC/NCEPLIBS-bufr/pull/519

### LINKED ISSUE #522: new release v12.0.1
What needs to be done:

- increment number in VERSION file
- update release notes, in both GitHub and in ReleaseNotes.md
- generate docs and update gh-pages branch with the new html, after first moving the existing html to previous_versions/v12.0.0

jbathegit:
Fixes #516
Fixes #521
Fixes #522


AlexanderRichert-NOAA (approved):

docs/user_guide.md
```diff
@@ -2,6 +2,7 @@
 
 ## Documentation for Previous Versions of NCEPLIBS-bufr
 
+* [NCEPLIBS-bufr-12.0.0](https://noaa-emc.github.io/NCEPLIBS-bufr/previous_versions/v12.0.0/index.html)
```
AlexanderRichert-NOAA, line 5: I assume this URL will exist after the 12.0.1 release?

jbathegit: Yes, once I officially release it and install the documentation (which just got generated in the Developer CI build ;-)

PR #529, opened 2023-10-07
Build pdoc

### LINKED ISSUE #230: update Python documentation via pdoc

AlexanderRichert-NOAA:
This PR enables python documentation through pdoc in the developer CI. It gets placed under the rest of the docs directory structure so it's in the same output artifact.

One caveat is that I've put the doc build in a separate CI step from the rest of the testing. This is because the use of the memory sanitizer and gcov don't play nicely with python. pdoc, for whatever reason, needs to load the python module, but when asan and gcov are enabled they put a bunch of special functions in the bufr library, so the python module load fails with "undefined reference" errors. I think the easiest solution is to put them in separate steps (as opposed to trying to hack together some LD_PRELOAD magic).

Fixes #230

### DIFF
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -40,6 +40,13 @@ jobs:
         path: ~/data
         key: data-10
 
+    - name: build-docs
+      run: |
+        mkdir bufr/build-docs
+        cd bufr/build-docs
+        cmake -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_BUILD_TYPE=Debug -DENABLE_DOCS=ON -DENABLE_PYTHON=ON ..
+        make doc python_docs
+
     - name: build
       run: |
         cd bufr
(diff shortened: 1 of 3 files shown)


jbathegit:
This looks great Alex - thanks!!

jbathegit (approved):

PR #534, opened 2023-11-02
Jack Woollen's fix to optimize upb8.f routine

jbathegit:
Courtesy of @jack-woollen, and per the discussions in [GSI/589](https://github.com/NOAA-EMC/GSI/issues/589) and [GSI/642](https://github.com/NOAA-EMC/GSI/issues/589), this patch to subroutine upb8 optimizes performance for the majority of values which are encoded in fields of 32 bits or less.

### DIFF
--- a/src/upb8.f
+++ b/src/upb8.f
@@ -36,16 +36,19 @@ C> @author J. Woollen @date 2022-05-06
 !----------------------------------------------------------------------
 !----------------------------------------------------------------------
 
-      if(nbits<0 ) call bort('BUFRLIB: UPB8 - nbits < zero !!!!!')
-      if(nbits>64) then
+      if(nbits<0) then
+         call bort('BUFRLIB: UPB8 - nbits < zero !!!!!')
+      elseif(nbits<=32) then
+         jbit=ibit; ival=0
+         call upb(ival,nbits,ibay,jbit)
+         nval=ival
+      elseif(nbits<=64) then
+         jbit=ibit; nvals=0
+         call upb(nvals(2),max(nbits-nbitw,0),ibay,jbit)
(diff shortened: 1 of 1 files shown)


jbathegit:
BTW, I saw in the discussions how there's more to come w.r.t. improving the GSI timings, but in the meantime I don't see any reason not to go ahead and get this into the library baseline, since it's certainly a useful update :-)

jack-woollen:
Here's an interesting fact I stumbled on. As you know version 11.7.1 has the same ability to pack up numbers in 64 bits as v12.0.0. When I test 11.7.0, 11.7.1, and 12.0.0 against each other, both 11.7.1 and 12.0.0 show the same time increase against 11.7.0. Which means the imb8 blocks have nothing to do with the slowdown. It narrows the focus to what's slowing down 11.7.1, which is a much smaller change (I think) than 12.0.0. I can approve this if you want but there may be another change or two worth making sometime soon.

jbathegit:
Yeah, that makes perfect sense Jack, given that the first issue you found was this one in upb8, and version 11.7.1 was indeed the first library release which included the capability to encode values larger than 32 bits (as noted in the [Release Notes](https://noaa-emc.github.io/NCEPLIBS-bufr/md__home_runner_work_NCEPLIBS_bufr_NCEPLIBS_bufr_bufr_docs_ReleaseNotes.html) ;-)

So that was a good catch, and yes that should considerably narrow down the scope of the search!

In the meantime, yes please approve this so we can go ahead and at least get this upb8 upgrade into the code baseline, but I completely understand that there's likely more to come and which can always be added via one or more additional PRs at a later date.

jack-woollen (approved):

PR #542, opened 2023-11-30
CI fixes

AlexanderRichert-NOAA:
This PR fixes the Intel compiler issue, and has a workaround for the macOS issue, and should at least let those CI workflows run through, with a more permanent fix for the macOS (hopefully) soon. IntelLLVM support (#536) may take some time as well while I'm waitiing to hear back from Intel devs.


jbathegit:
Thanks for this @AlexanderRichert-NOAA - much appreciated!

jbathegit (approved):

Issue #533, opened 2023-11-01
debufr failing with internal write overflow


DavidHuber-NOAA:
When compiled with spack-stack on Hera, the debufr application is failing with an internal write error:

```
[David.Huber@hfe01 .github]$ module use /scratch1/NCEPDEV/nems/role.epic/spack-stack/spack-stack-1.5.1/envs/unified-env/install/modulefiles/Core
[David.Huber@hfe01 .github]$ module load stack-intel/2021.5.0
[David.Huber@hfe01 .github]$ module load bufr/12.0.1
[David.Huber@hfe01 .github]$ debufr /scratch1/NCEPDEV/nems/David.Huber/GSI/ptmp/tmpreg_global_4dvar/global_4dvar_loproc_updat/nsstbufr -o tmp
forrtl: severe (66): output statement overflows record, unit -5, file Internal Formatted Write
Image              PC                Routine            Line        Source
libifcoremt.so.5   00007F9D5FED9E79  for__io_return        Unknown  Unknown
libifcoremt.so.5   00007F9D5FF3BCAE  for_write_int_fmt     Unknown  Unknown
libbufr_4.so       00007F9D5FA3105A  mtfnam_                    57  mtfnam.f
libbufr_4.so       00007F9D5FA74E90  ireadmt_                  213  ireadmt.f
libbufr_4.so       00007F9D5FA24978  getcfmng_                 151  getcfmng.f
debufr             000000000040F8D3  fdebufr_f                 261  debufr.F90
debufr             000000000040E0CB  main                      346  debufr.c
libc-2.17.so       00007F9D5EAF5555  __libc_start_main     Unknown  Unknown
debufr             000000000040D329  Unknown               Unknown  Unknown
```

When compiled with hpc-stack, there is no issue.  Looking into mtfnam.f and ireadmt.f, it appears that the maximum length for a BUFR table filename is 132 while for the Hera installation, the full path length for the tables is 155-156.  This should probably be increased.

FYI @wx20jjung

jbathegit:
It looks like we do have a bit of an inconsistency in the library, because variable MTDIR is dimensioned as character\*240 in mtinfo.f, but only as character\*100 in mtfnam.f and ireadmt.f.  So we should probably bump the latter two up to character\*240 (which would also match the length of cmtdir in bfrini.f90.in), and then correspondingly bump STDFIL and LOCFIL to be character\*275 in ireadmt.f

[...]

PR #543, opened 2023-12-01
Jack Woollen's fix to optimize rdcmps.f routine

jbathegit:
Courtesy of @jack-woollen, and per the discussions in [GSI/642](https://github.com/NOAA-EMC/GSI/issues/642), this patch to subroutine rdcmps optimizes performance for the majority of values which are encoded in fields of 32 bits or less.

### DIFF
--- a/.github/workflows/Intel.yml
+++ b/.github/workflows/Intel.yml
@@ -50,7 +50,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/data
-        key: data-10
+        key: data-11
 
     - name: build
       run: |
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -40,7 +40,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/data
-        key: data-10
+        key: data-11
 
     - name: build
       run: |
--- a/.github/workflows/MacOS.yml
+++ b/.github/workflows/MacOS.yml
@@ -44,7 +44,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/data
-        key: data-10
+        key: data-11
 
     - name: build-bufr
       run: |
--- a/.github/workflows/developer.yml
+++ b/.github/workflows/developer.yml
@@ -38,7 +38,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/data
-        key: data-10
+        key: data-11
 
     - name: build-docs
       run: |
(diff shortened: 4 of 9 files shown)


jbathegit:
Thanks again @jack-woollen for this fix, but before we can merge it we'll need to add a test case to the CI test suite, because right now the lines of code in the new `else if (nbit<=64)` branch are untested.  So we'll basically need a compressed BUFR message which has at least one encoded value of more than 32 bits.

If you don't know of an example off the top of your head, then I can work to try to gin one up when I get a chance, but I may not get to this right away b/c I'll be busy with other stuff in the next few weeks up to the holiday break.

In fact, and come to think of it, we're going to need to update the emcrzdm tar archive of test cases anyway as part of an eventual solution for #533, so maybe we can combine both new test cases into the same archive update.

jack-woollen:
Turns out 11.7.1 is the default on hera in hpc-stack:

 /scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/compiler/intel/18.0.5.274 ------------------------------------------------------------------------------------------------------
   bufr/11.4.0    bufr/11.5.0anaconda    bufr/11.5.0    bufr/11.6.0    bufr/11.7.0    bufr/11.7.1 (D)

jbathegit:
I've been working on generating a compressed BUFR message larger than 32 bits that we can add as a new test case in test_debufr.sh.  However, while doing so, I uncovered a bug in subroutine wrcmps, as shown in this most recent commit.

jack-woollen:
@jbathegit I don't know how I marked it ready for review as it says above. Let me know when it is and I will review it.

jbathegit:
You must have accidentally clicked the button to change it from a draft pull request to a regular pull request, which by definition means it's ready for review.

However, it won't really be ready until I can push up the aforementioned test case to the emcrzdm archive and correspondingly modify test_debufr.sh.  But before I modify the archive by pushing up the new test case, I also want to look at #533 as well. 

Bottom line - no worries, and I'll let you know once this PR is really ready for review.  Just FYI that I may not get all of the above done until after the holidays, b/c I'll be away at AGU next week.

jbathegit:
OK, I had an unexpected block of time to work on this, so I've now added a test case to test_debufr.sh and renumbered the cases behind it within that same script.

@jack-woollen this PR is now ready for review, and I'd definitely appreciate if you have some time to now look this over and approve it, because I also have a fix for #533 ready to go right behind it :-)

jack-woollen (approved):

PR #548, opened 2023-12-20
more testing

jbathegit:
Part of #445


jbathegit:
Note that I removed the `TAG(NODE)` references from the calls to bort() within lstjpb(), because if it reaches those statements then we already know that the value of NODE is out-of-bounds within lstjpb(), and therefore it makes no sense to try to dereference that member of the TAG array.  In fact, doing so will often end up segfaulting.

jbathegit:
I added some testing for the print lines in ireadmt() at verbosity level 2, and I also cleaned up a print statement in mtfnam().

The calls to bort() within ireadmt() are going to be hard to test, because these only get triggered if either the STD or LOC file exists on the filesystem but for some reason can't be opened.  If either file simply doesn't exist on the filesystem, then that error gets trapped in the earlier call to mtfnam(), and both of those cases are already being tested.

jbathegit:
Check out the latest GCC numbers :-)

NCEPLIBS-bufr gcc

jack-woollen (approved):

PR #551, opened 2024-01-22
more testing

jbathegit:
Part of #445


jbathegit:
FYI, the bort() 904 and 905 tests have been removed from ufbseq() because they're unreachable.  If ufbseq() is called with either I1 or I2 set to a non-positive number, then this gets trapped as a warning and returns with IRET = 0 earlier on in the routine.  So there's no way it can ever get to those bort() statements later on in the routine.

jbathegit:
Further testing also showed that the first of the two bort() 900 lines in getntbe() was unreachable.

In getntbe(), any LINE returned from a successful call to igetntbl() is guaranteed to be a non-empty string.  Therefore, whenever that LINE is subsequently passed into parstr(), we're guaranteed that the return value NTAG will always be at least 1.

edwardhartnett:
Awesome to see all the testing, and great that you are discovering unreachable code paths which can be removed, lightening the maintenance burden and making the code more clear.

jbathegit:
OK I think that's enough for now :-)

Overall test coverage is now up to 92.3% 

gcovr

jack-woollen (approved):

edwardhartnett:
Well the code coverage report has added a new color! Yellow!

image

Looks like the yellow lines are for conditionals where only one of the conditions has been tested. 

(BTW this seems like it would be an easy function to get to full testing...)

PR #553, opened 2024-01-31
still more testing

jbathegit:
Part of #445

FYI, the RETURN statement at the end of isize() was unreachable and has been removed, and ditto for the BORT 903 case in ufbqcd(), since any string returned from adn30() is guaranteed to only contain numeric characters.


jbathegit:
I also removed the BORT 900 and 901 cases from cpymem(), rdmemm(), and rdmems() because they're unreachable.

In all three of these routines, the prior call to status() is checking the logical unit MUNIT, which is not passed in directly but rather is set internally in the library via a previous call to ufbmem() or ufbmex().  But if an application previously calls either of those routines, then logical unit MUNIT is already guaranteed to be open for input, so there's no way to trigger either of those BORT cases.  Conversely, if neither ufbmem() or ufbmex() has been previously called by an application, then MUNIT is set internally to a default value of 0, and in which case a separate BORT will get triggered within the prior call to status().  So again, there's no way to trigger either of those BORT cases in any of the above three routines.

AlexanderRichert-NOAA (approved):

jack-woollen (approved):

Issue #328, opened 2023-02-18
use implicit none in all Fortran codes


edwardhartnett:
implicit none guards against one of the most common programming bugs: mispelling a varaible.

When we misspell a fortran variable, and a new one is created, it's extremely easy to introduce a bug. For example:

```
C           Store this message into MODULE MSGMEM.

            ICT = ICT + 1
            IF ( ( NDXM + ICT ) .GT. MXDXM ) GOTO 902
            IPDXM(NDXM+ICT) = LDXM + 1
            LMEM = NMWRD(MGWA)
            IF ( ( LDXM + LMEM ) .GT. MXDYW ) GOTO 903
            DO J = 1, LMEM
              MDX(LDXM+J) = MGWA(J)
            ENDDO
            LDXM = LDXM + LMEM

```
Where is the bug in this code? If it's obvious to you, it certainly is not to me, nor will it be to most programmers.

The good news is that the Fortran compiler will tell us exactly where the bug is, and will refuse to let us build the code at all, until we fix it. All we need to do is use implicit none to get this fantastic, automatic, and infallible bug finder.

edwardhartnett:
I think we all agree that implicit none should be used in all new code (including tests). We will add implicit none to existing code after full testing, and as we refactor or change the code.

edwardhartnett:
@jbathegit are we adding implicit none to all codes? When I look at the code I don't see them, even in subroutines where it would be really easy to put it there.

Do you not agree with this issue?

jbathegit:
I do agree that this is a worthwhile goal, mostly for the reason you noted of more quickly finding a misspelled variable.  I just haven't done any further work on this, because I've been prioritizing whatever time I have to work with NCEPLIBS-bufr on getting us to a more full level of testing, as you yourself noted above.

(And BTW, the bug in the above snippet is the misspelling MXDYW, which should be MXDXW :-)

PR #556, opened 2024-02-05
consolidate all global vars into a single module

### LINKED ISSUE #430: why not one module for all module vars?
We have:

```
!> This module declares and initializes the BMISS variable.
!>
!> @author J. Ator @date 2021-03-24
module modv_bmiss
  !> Current placeholder value to represent "missing" data when reading
  !> from or writing to BUFR files; this value can be changed at any
  !> time via a call to subroutine setbmiss().
  real*8, public :: BMISS = 10E10_8
end module modv_bmiss

!> This module declares and initializes the IFOPBF variable.
!>
!> @author J. Ator @date 2015-03-03
module modv_ifopbf
  !> Status indicator to keep track of whether subroutine openbf() has
  !> already been called:
  !> - 0 = No
  !> - 1 = Yes
  integer, public :: IFOPBF = 0
end module modv_ifopbf

!> This module declares and initializes the IM8B variable.
!>
!> @author J. Woollen @date 2022-08-04
module modv_im8b
  !> Status indicator to keep track of whether all future calls to
  !> BUFRLIB subroutines and functions from a Fortran application
  !> program will be made using 8-byte integer arguments.
  !>
  !> The default value is .false., meaning that all future calls to
  !> BUFRLIB subroutines and functions will be made using 4-byte
  !> integer arguments.  This value can be changed at any time via a
  !> call to subroutine setim8b().
  logical, public :: IM8B = .false.
end module modv_im8b


```

etc.

Seems like it would be natural to have one module, moda_vars, and have all the vars in the same module. This would reduce our module count quite a bit. Also subprograms would only have to have one use statement to get all the variables, instead of one use statement for each variable, as is currently the case.

This would also simplify the documentation, which currently has a module for each of these variables, which is a bit much.

jbathegit:
Fixes #430


edwardhartnett (approved):
Nice.

I'm glad we have a lot of testing to ensure there were no typos...

jbathegit:
> I'm glad we have a lot of testing to ensure there were no typos...

Yeah I was thinking the same thing as I was doing this.  Another reason to make some progress on #328 where/when we can.

PR #558, opened 2024-02-06
more tests

jbathegit:
Part of #445

### DIFF
--- a/src/modules_arrs.F90
+++ b/src/modules_arrs.F90
@@ -96,7 +96,7 @@ end module moda_bufrmg
 !> @author J. Ator @date 2014-12-10
 module moda_bufrsr
   !> File ID of BUFR file.
-  integer :: JUNN
+  integer :: JUNN = 0
   !> File status indicator of BUFR file.
   integer :: JILL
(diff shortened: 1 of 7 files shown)


jbathegit:
FYI, the BORT 906 statement in readlc() is unreachable, because MSGUNP(LUN) can never have any value other than 0, 1, or 2.  So after taking out that ELSE branch, I also simplified the existing ELSEIF(MSGUNP(LUN).EQ.2) THEN statement to just be an ELSE statement.

jbathegit:
FYI, rewnbf() in it's previous form would core dump if a user mistakenly called the routine with ISR=1 without having previously called it with ISR=0.  In that case, the value of JUNN would be undefined, so the evaluation of JSR(JUNN) would result in a core dump.  I fixed this by explicitly initializing JUNN=0 and then adding a second BORT 902 test to check for that value.

jbathegit:
This PR now gets us up to 93.3% line coverage and 71.6% branch coverage.

AlexanderRichert-NOAA (approved):

PR #560, opened 2024-02-07
use standard F2003 argument functions in utils

### LINKED ISSUE #559: Undeclared iargc
I am getting failures building with some compilers of this sort:

ftn-232 ftn: ERROR SPLIT_BY_SUBSET, File = ../../../../../lustre/research/scratch/d01/david.davies/cylc-run/EXZCrayPorting/work/1/build_ncep_bufr__exz_cray/src/utils/split_by_subset.f90, Line = 34, Column = 6 
  IMPLICIT NONE is specified in the local scope, therefore an explicit type must be specified for function "IARGC".

This is from the Cray compiler but there are other compilers that fail in a similar way.

jbathegit:
Fixes #559

### DIFF
--- a/utils/binv.f90
+++ b/utils/binv.f90
@@ -27,13 +27,13 @@ PROGRAM BINV
 
   !  get filename
 
-  NARG=IARGC()
-  IF(NARG/=1) THEN
-     PRINT *,'Usage: binv <bufrfile> will print bufrfile inventory by message type'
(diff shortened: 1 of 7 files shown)


edwardhartnett:
Looks like I better do this for NCEPLIBS-grib_util as well!

jack-woollen (approved):

PR #563, opened 2024-02-13
remove wrdlen routine and common blocks /hrdwrd/ and /erdord/

### LINKED ISSUE #382: subprogram WRDLEN() should be removed, and replace by an inc file filled in by CMake and configure time
We have a subroutine WRDLEN(). From the documetation:

```
C> This subroutine figures out some important information about the
C> local machine on which the BUFRLIB software is being run, including
C> the native endianness, the number of bytes in an integer, and
C> whether the machine uses the ASCII or EBCDIC character set.

```

This is the kind of thing best done in CMake. CMake has all the tools to support this easily, and set up a .inc file that can be included by the Fortran code which will have all the answers.

The benefits are:
* Done once per machine instead of every library use.
* Removes Fortran code for less CMake code - cmake already supports finding system stuff without us writing code.
* Allows us to switch to using pre-processor symbols for some of these, which may be appropriate since these represent fundamental machine/compiler properties such as the endianneess.

jbathegit:
Fixes #382 

Part of #254

This PR also simplifies the logic in the blocks routine and removes a large chunk of unreachable code from the openbf routine.


jack-woollen (approved):

edwardhartnett:
Great to see these improvements and modernizations...

PR #565, opened 2024-02-13
Added CmpMsg to python interface

### LINKED ISSUE #562: Python and BUFR compression.
The python bindings don't seem to to support writing compressed BUFR data (ex: cmpmsg.f is not available). Would be nice to have this. My use case is that I want to generate some test data by sampling some larger BUFR files into much smaller representative data sets...

rmclaren:
Expose the cmpmsg function via the python interface so you can cause written data to be compressed.

Fixes https://github.com/NOAA-EMC/NCEPLIBS-bufr/issues/562


jbathegit:
@rmclaren did you test this out and confirm it works like you expected?

The reason I'm asking is b/c of your line `return _bufrlib.cmpmsg(cmp)`   Specifically, cmpmsg is a subroutine, not a function.  Therefore, it has no return value, so what exactly would you be "returning" from within the corresponding python binding?

jbathegit:
@rmclaren do you still want this hook into cmpmsg() added to the Python interface?  If so, it has no return value, so it needs to be called as a subroutine instead of as a function.

Let me know if you still want this feature, and I can push up another commit to get this detail fixed and then merge it.

rmclaren:
@jbathegit I don't really "need" it, it was just a nice to have for a one-off script I was writing... I made the change you requested, but it doesn't really change anything. Any python function will implicitly return None (void) whether explicitly call return or not.

jbathegit (approved):

PR #566, opened 2024-02-14
convert some routines to .F90

jbathegit:
Part of #254
Part of #328


src/pkc.F90
```diff
@@ -0,0 +1,90 @@
+!> @file
+!> @brief Encode a character string within an integer array.
+!>
+!> @author J. Woollen @date 1994-01-06
+
+!> Encode a character string within a specified number of bytes of an integer array,
```
edwardhartnett, line 6: Did you delete and readd the files? Or use git mv?

If you used git mv I would expect to be able to see any changes in the code, but the review shows all the code as new...

jbathegit: I always use "git mv" for this sort of thing, and it displays like you would expect when you do a "git diff" locally, but then what it uploads to GitHub it ends up looking like you're seeing it now.  I'm not sure why it does that, but yes I did use "git mv".

jbathegit: It looks like it may actually be happening during the "git commit" on the local machine, rather than via GitHub.  For example, if I do a "git status" I see the following, which is what I'd expect:

Capture1

But then when I do a "git commit" I see:

Capture3

In this case I didn't make any edits to any of the utils codes so those still look like they did before.  But I did have to edit the src files to convert them from f (fixed-form) to F90 (free-form), and now those look like they were actually deleted and created as new files!?

jbathegit: At any rate I'll go ahead now and mark this as Ready for Review to keep the size of the PR manageable

jbathegit:
Looking at the latest gcovr report, I found one more thing I could clean up.

Specifically, since I made the recent change to use iupbs01() instead of iupb() inside of the xbfmg utility, then the latter routine is no longer being called from anywhere in any C code.  So I was able to remove the iupb_c() function and its associated iupb_f wrapper.

jack-woollen (approved):

src/pkc.F90
edwardhartnett: Yes, its fine I was just curious...

PR #569, opened 2024-02-20
yet even more testing

### LINKED ISSUE #349: complete testing of copysb.f

jbathegit:
Fixes #349
Part of #445

### DIFF
--- a/.github/workflows/Intel.yml
+++ b/.github/workflows/Intel.yml
@@ -50,7 +50,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/data
-        key: data-13
+        key: data-1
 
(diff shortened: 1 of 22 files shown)


jbathegit:
The bort 907 statement in copysb() was unreachable, because icmp can only ever be 0 or 1 when mesgbc() is called with a negative first argument.

I also found I couldn't practically test the bort 900 statement in reads3() when MXCNEM was fixed as a parameter at a value of 450.  So I went ahead and added it to the set of global variables that can be specified at runtime via isetprm().

jack-woollen (approved):

PR #571, opened 2024-02-22
still more tests

jbathegit:
Part of #445

FYI, I removed unreachable lines from ufbrep() and ufbstp().  In both of those routines, whenever IO=1 and either ufbrp() or ufbsp() returns with IRET=0 (or any other value not equal to I2), then control immediately drops to the bort 903 message, bypassing those print lines.

Similarly, in ufbovr() the same print lines are unreachable because any situation where IO=1 and IRET!=I2 will end up aborting from lstjpb() within the prior call to trybump().  So in the case of ufbovr(), I'm not sure if the bort 904 case is itself even reachable, but for now at least I left it in.

I also simplified the code by removing the global variable MAXTAB, since it's nothing more than a copy of MAXJL.

### DIFF
--- a/src/bfrini.F90.in
+++ b/src/bfrini.F90.in
@@ -14,7 +14,7 @@
 !> @authors J. Woollen J. Ator @date 1994-01-06
 subroutine bfrini
 
-use modv_vars, only: maxjl, maxtba, maxtbb, maxtbd, mxmsgl, nfiles
+use modv_vars, only: maxtba, maxtbb, maxtbd, mxmsgl, nfiles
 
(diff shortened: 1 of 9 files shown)


jbathegit:
The latest numbers...
Capture

jbathegit:
Hi @jack-woollen and @edwardhartnett, would one of you mind taking a look at this and let me know if you have any questions/concerns?  Thanks, b/c if possible I'd like to get it merged before moving on to other stuff.

jack-woollen (approved):

PR #573, opened 2024-02-26
consolidate bort and errwrt logic, and convert to F90

### LINKED ISSUE #373: combine files bort.f, bort2.f, bort_exit.f, errwrt.f into error_handling.F90
I suggest we combine files bort.f, bort2.f, bort_exit.f, errwrt.f into one new F90 file called error_handling.F90.

This should only be done after we have tested all our error handling code.

Any other error handling subroutines should go in error_handling.F90, it should use free-format (i.e F90 style formatting and line continuation.)

All new error handling subroutines would then also go in this file.

jbathegit:
Fixes #373

Part of #254
Part of #328


jbathegit:
I was able to eliminate bort_exit.c, and I also consolidated bort.f and bort2.f into a single new F90 file.  But I couldn't consolidate errwrt into the same F90 file b/c of a naming conflict.

Specifically, errwrt is basically a placeholder that just prints error strings to stdout by default, but the intent is that application codes can define their own version of errwrt to override the default library version if they want those strings to go someplace besides stdout (such as to syslogd, or a decoder log, etc.).  In the past, including with the many test codes which have their own version of errwrt, we've never experienced a conflict, and if the application program defined its own version of errwrt then the compiler would always use it and correctly override the default version in the library.  However, and for whatever reason, if I don't keep errwrt in its own separate F90 file, the GNU compiler squawks with a naming conflict and apparently can't figure out for itself which one it's supposed to use.  So after much trial and error, I ended up just keeping errwrt in its own new F90 file, and the net result is that I was able to consolidate 4 previous source files into 2 new source files, rather than all the way down into 1 new source file.  But still an improvement.

jack-woollen (approved):

PR #576, opened 2024-03-06
add testing for igetrfel, rcstpl, and pad

jbathegit:
Part of #445 

Also moves the definition of parameter maxrcr from modules_arrs.F90 to modules_vars.F90.  This had been the last remaining global variable to not be defined in the latter.

I also removed some unreachable bort statements.

### DIFF
--- a/src/arallocf.F90
+++ b/src/arallocf.F90
@@ -17,7 +17,7 @@ subroutine arallocf
 
   use modv_vars, only: maxcd, maxjl, maxmem, maxmsg, maxss, maxtba, maxtbb, maxtbd, mxbtm, mxbtmse, &
                        mxcdv, mxcsb, mxdxts, mxlcc, mxmsgl, mxmsgld4, mxmtbb, mxmtbd, mxnrv, mxrst, &
-                       mxs01v, mxtamc, mxtco, mxh4wlc, nfiles, mxcnem, maxnc
+                       mxs01v, mxtamc, mxtco, mxh4wlc, nfiles, mxcnem, maxnc, maxrcr
 
   use moda_usrint
   use moda_usrbit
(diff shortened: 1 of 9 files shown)


jbathegit:
The bort 900 statement in pad() was a particularly interesting case to try and decipher, so I'm going to document it here for the record in case there's any future questions about it.

To begin with, ibit is both an input and output parameter, and ipadb is an input parameter.  The logic in the routine takes the existing value of ibit and advances it first by 8 bits (to store the number of padded bits as a delayed replication factor) and then by another (ipadb - mod(ibit+8,ipadb)) bits (to store the actual padded bits as all zeroes).

So, in a nutshell, the final value of ibit (let's call it ibit2) is `ibit2 = ibit + 8 + ipadb - mod(ibit+8,ipadb)`, and then the subsequent test is to bort if `mod(ibit2,ipadb) .ne. 0` .  Or in other words, to bort if ibit2 isn't a multiple of ipadb.  And this is what it seems to me can never happen.  If we set X = (ibit+8) and Y = ipadb, then the above equation becomes a simpler question of whether `mod(X + Y - mod(X,Y), Y)` can ever have a remainder for any integers X and Y.  To answer that question, consider that there are 3 possibilities for mod(X,Y):

1.  if X=Y, then mod(X,Y) = 0, and the equation `mod(X + Y - mod(X,Y), Y)` in turn reduces to `mod(2*Y,Y)`, which of course is 0.
2. if X<Y, then mod(X,Y) = X, and the equation `mod(X + Y - mod(X,Y), Y)` in turn reduces to `mod(Y,Y)`, which is also 0.
3. if X>Y, then mod(X,Y) = (X-(N*Y)), where N is some positive integer >=1, and the equation `mod(X + Y - mod(X,Y), Y)` in turn reduces to `mod(X + Y - (X - (N*Y)),Y)`, and then further to `mod((N+1)*Y,Y)`, which is also 0.

So bottom line, the previous bort 900 test in pad() was unreachable, unless I messed up something in my analysis :-)

jbathegit:
BTW, this latest PR now puts us at 95.3% testing overall for the library :-)

jack-woollen (approved):

PR #583, opened 2024-04-08
more F90 updates

jbathegit:
Part of #254
Part of #328


src/readwritesb.F90
```diff
@@ -0,0 +1,598 @@
...
+!> arrays within the NCEPLIBS-bufr software, and from where it can be
+!> manipulated or further parsed via calls to any of the [values-reading
+!> subroutines](@ref hierarchy).
+!>
+!> If the subroutine returns with iret = -1 there are no more data
+!> subsets available within the current message; a new call needs to
+!> be made to one of the [message-reading subroutines](@ref hierarchy)
+!> in order to read in the next message from logical unit lunit.
+!>
+!> @param[in] lunit - integer: Fortran logical unit number for BUFR file
```
edwardhartnett: One thing you might notice is that with F90 code, doxygen can find the types of parameters. That is, you no longer need to have:
`@param[in] lunit - integer: Fortran logical unit number for BUFR file
`
You could switch to:
`@param[in] lunit - Fortran logical unit number for BUFR file
`

jbathegit: OK, I didn't know that.  But if that's the case then I'll start leaving those extra descriptions out to avoid redundancy - thanks!

jbathegit:
I know this looks like a lot, but of all the "Files changed", half of them are files being deleted, plus I've also gone back and simplified all of the variable declarations in all of the F90 files per Ed's earlier suggestion :-)

AlexanderRichert-NOAA (approved):

PR #599, opened 2024-06-03
more updates

jbathegit:
Part of #254
Part of #328
Part of #445


jbathegit:
OK all good now.  There was a bug in rewnbf that was always restoring the subset pointer to point to the first subset (i.e. the beginning) of a restored BUFR message whenever the subroutine was called with isr=1, even if the subset pointer had been pointing to a different subset in the message when the routine was called previously with isr=0.  I had fixed this bug in an earlier commit, but there was an existing Python test which had been overcoming this bug by re-reading up to the expected subset in the message.  So once I fixed the bug in the library, that Python test began failing until I now just removed the extra script lines which had previously been re-positioning the subset pointer to overcome the bug.

jack-woollen (approved):

PR #603, opened 2024-06-14
switch to newer relational operators

jbathegit:
As an addendum to the recent F90 modernization, I've also gone through now and switched all of the relational operators to the newer versions, which I think makes the code easier to read.  So basically switching all .eq. to ==, .ne. to /=, .gt. to >, .le. to <=, etc.

This was super-easy once I cooked up a series of perl commands to do this for multiple files in a directory simultaneously.  I'll share them here in case anyone else wants to use them in other libraries or codes:

```
perl -pi -e 's/\.eq\./==/g' *F90
perl -pi -e 's"\.ne\."/="g' *F90
perl -pi -e 's/\.ge\./>=/g' *F90
perl -pi -e 's/\.le\./<=/g' *F90
perl -pi -e 's/\.gt\./>/g' *F90
perl -pi -e 's/\.lt\./</g' *F90
```


AlexanderRichert-NOAA:
Looks good to me- I don't have a strong preference either way on which style of comparison operators to use. I haven't looked at the F90 autoformatting stuff for a while that I was experimenting with on the ip lib, but if we do go that route, this is a standard that could be uniformly applied across NCEPLIBS.

AlexanderRichert-NOAA (approved):

PR #604, opened 2024-06-14
add new -DBUILD_UTILS option to cmake

### LINKED ISSUE #378: add option to turn off building and testing of utils
We should have an option BUILD_UTILS, ON by default. When turned off the utils should not be built, tested, or installed.

Most data format packages have this feature, in my experience, including HDF4, HDF5, and netCDF.

For example, in any NCEPLIBS CI run, we would not need the utilities to meet a dependency on NCEPLIBS-bufr.

jbathegit:
Fixes #378


jbathegit:
Hey @AlexanderRichert-NOAA, the Spack/recipe-check test is now failing b/c of the new -DBUILD_UTILS option:

Capture


I'm not sure what the best default would be to set this to in spack/package.py, so would you please take a look at this?

And BTW, we already have a separate -DBUILD_TESTING option available in the library, so how exactly does that relate to the BUILD_TESTS option that's also defined in spack/package.py?  Or are they two separate settings for different purposes?  Again, BUILD_TESTING is what is used throughout the rest (i.e. non-Python parts) of the library :-)

AlexanderRichert-NOAA:
I added a variant corresponding with BUILD_UTILS to the Spack recipe. See my PR https://github.com/NOAA-EMC/NCEPLIBS-bufr/pull/605 that points to your jba_utilsoption branch.

As for "BUILD_TESTS," it is in the recipe because that's the option was called in older versions; with the newer versions it just gets ignored.

jbathegit:
Got it - thanks Alex!

jbathegit:
Thanks again @AlexanderRichert-NOAA for your help to resolve this!

When you get a chance, could you (or @edwardhartnett or @jack-woollen) please review and approve the rest of this PR so we can get it merged?

AlexanderRichert-NOAA (approved):

Issue #586, opened 2024-04-09
new release v12.1.0


jbathegit:
What needs to be done:

- increment number in VERSION file
- update release notes, in both GitHub and in ReleaseNotes.md
- move the existing html to previous_versions/v12.0.1, and add a new reference link for it within user_guide.md
- generate docs and update gh-pages branch with the new html

jbathegit:
FYI, the VERSION file has already been incremented to 12.1.0 in the develop branch, re: #554

jbathegit:
For this specific release, also update README.MD with the following:

- In the directive "_**Both path1 and path2 may be full or relative pathnames on the system, up to a maximum of 90 characters each.**_", change _**90**_ to _**240**_.
- Add a new directive "_**When building the library, associated utilities are included by default.  If these are not desired, then they can be disabled by adding `-DBUILD_UTILS=OFF` to the above cmake command.**_"
- Add a new directive "_**When building the library, automated CI testing is included by default.  If this is not desired, then it can be disabled by adding `-DBUILD_TESTING=OFF` to the above cmake command.**_"

jbathegit:
Note that we may also need some new "12.1.0" entries in spack/package.py, but how do we get the corresponding sha256 hash code before actually creating the tag? In other words, is this a chicken vs. egg question?

jbathegit:
> Note that we may also need some new "12.1.0" entries in spack/package.py, but how do we get the corresponding sha256 hash code before actually creating the tag? In other words, is this a chicken vs. egg question?

Never mind, I found the answer to this myself in #552 and made the subsequent updates in #609

jbathegit:
After making any future releases, also do the following:

1. Notify Alex Richert, Ed Hartnett, and Hang Lei for updating spack-stack
2. Open a new issue at https://github.com/NOAA-EMC/WCOSS2-requests for installation on WCOSS2

PR #606, opened 2024-07-08
updating tables directory with version 42 of WMO master tables

### LINKED ISSUE #584: add new version 42 updates to WMO master table
WMO releases new versions of their master BUFR table every 6 months, and these in turn need to be reflected in updates to the tables subdirectory of the NCEPLIBS-bufr project so that the software can read any BUFR message encoded using the new version of the table. The newest version of the WMO master BUFR table is version 42, effective mid-May 2024, so we need corresponding new NCEPLIBS-bufr v42 master BUFR table files for TableB, TableD and CodeFlag.

jbathegit:
Fixes #584


jbathegit:
Once this is merged, my next task will be #586 to do the next v12.1.0 release :-)

jbathegit:
BTW, and I guess this may be a question for @AlexanderRichert-NOAA, but there's still a pending question in the thread for #586 which I'm unclear about; specifically, for the next release it looks like we may also need some new "12.1.0" entries in spack/package.py, but how do we get the corresponding sha256 hash code before actually creating the tag? In other words, is this a chicken vs. egg question?

jack-woollen (approved):

PR #613, opened 2024-08-02
add logic to encode Table C marker operators

### LINKED ISSUE #611: add capability to write/encode Table C marker operators
Table C operators such as 224255, 232255, etc. are fairly esoteric and have only been used in a handful of data types during the long history of BUFR which dates back 30+ years.  Nevertheless, they've still been a part of the official BUFR regulations (and therefore legal for international exchange) during much of that time.

NCEPLIBS-bufr has had the capability to read/decode data associated with these operators since the release of v11.2.0 back in April 2017.  However, it still lacks the capability to write/encode data associated with these operators.

jbathegit:
Fixes #611


jbathegit:
@AlexanderRichert-NOAA @edwardhartnett @GeorgeGayno-NOAA @Hang-Lei-NOAA 

FYI, and as of about a week ago, it looks like the MacOS runner no longer has gfortran-11 available to use as a Fortran ($FC) compiler option.  See [here](https://github.com/actions/runner-images/issues/10213) for more details.

So I just went ahead and updated to gfortran-12 in the MacOS workflow for NCEPLIBS-bufr.

jack-woollen (approved):

PR #629, opened 2024-10-25
CI updates

AlexanderRichert-NOAA:
This PR resolves a few CI issues. There seem to be some doxygen bugs if I update to the ubuntu-24.04 runner in developer.yml, so I'm leaving that set to 22.04. Linux.yml is/needs to be explicitly set to 24.04 in order to use the new pip flags. The others I'm leaving on ubuntu-latest because the exact version doesn't seem to matter, and GitHub Actions is still going back and forth between 22.04 and 24.04 for "ubuntu-latest"...

There were a lot of debugging commits, so I recommend squashing and merging.


AlexanderRichert-NOAA:
If/when everything is approved I can update the required tests based on the new gcc versions for Linux.yml.

jbathegit:
Thanks Alex, and I can definitely squash+merge when I do the merge into the develop branch.

**Just one remaining question**: it looks like Doxyfile.in has been updated to v1.9.8.  At first I presumed that was b/c the developer runner is now using that later version of Doxygen, but then when I looked into the details of the build-docs step in the developer run, I saw that it was still using v1.9.1 of Doxygen.  So why the update to v.1.9.8 in Doxyfile.in?  Is that just preparatory for an eventual move to Doxygen v1.9.8 once we're able to update to ubuntu-24.04 in the developer runner, and which you noted above you had to leave at ubuntu-22.04 for now?

AlexanderRichert-NOAA:
Yes-- I updated it to avoid doxygen complaining/erroring out when using ubuntu-24.04, but that version didn't work out anyway so it's not strictly necessary at the moment.

jbathegit (approved):

jbathegit:
OK it's approved.  Once you remove the Linux gcc-9 and gcc-10 tests, then I'll go ahead and squash+merge.

jbathegit:
> OK it's approved. Once you remove the Linux gcc-9 and gcc-10 tests, then I'll go ahead and squash+merge.

Or I can do this myself if you want.  I can also go into the branch protection settings and remove those tests (and add the gcc-12 and gcc-13 ones in their place :-)

AlexanderRichert-NOAA:
Okay it's good to go

jbathegit:
Thanks again Alex for your help in getting this resolved!

PR #630, opened 2024-10-25
increase maximum message size in debufr utility

jbathegit:
Also relocated an initialization loop to improve runtime efficiency, per discussion in #625 

Also cleaned up an address sanitizer issue that had been showing up in the developer runner for subroutine rewnbf


src/jumplink.F90
```diff
@@ -206,7 +206,9 @@ subroutine makestab
   ! Set up expansion segments for type 'SUB', 'DRP', and 'DRS' nodes.
 
   newn = 0
-
+  do k=1,maxjl
```
AlexanderRichert-NOAA, line 209: Would it make sense to assign this as `knt(1:maxjl) = 0`?

AlexanderRichert-NOAA:
Is there a corresponding issue for the message length change?

jbathegit:
> Is there a corresponding issue for the message length change?

No I didn't write one up b/c it was literally a one-line change, and it came about b/c of an email from a WCOSS2 user who was having trouble using debufr to read through a file.  As I looked into the issue it became clear it was because some of the messages the user was trying to read were upwards of 4Mb in size, and the utility previously had an upper limit of 2.5Mb.

src/jumplink.F90
jbathegit: Interesting - I hadn't seen that shorthand notation before, and in fact there are probably numerous places throughout the library where a similar change could be made.  I can look into that more as part of a future update to the library.

jbathegit:
Hi @AlexanderRichert-NOAA @jack-woollen @edwardhartnett 

Are there any remaining comments or concerns about this PR?  If not, could someone please approve it so I can go ahead and merge it?

(Thanks! :-)

edwardhartnett (approved):

PR #637, opened 2024-12-18
Various fixes for using NAG

DJDavies2:
I have made some changes to enable the use of NAG.  Changes are:

1)

The changes in mastertable.F90 are related to intent(out) arguments being used without being defined.  For instance in subroutine sntbde, nmtbd is declared as intent(out) and promptly used in a comparison.  Intent(out) arguments become undefined upon subroutine entry so this is undefined.

2)

In missing.F90 there is an attempt to write a real value with a Z edit descriptor.  This is not valid according to the standard, only integers can be used with a Z edit descript.  I have used a transfer to fix this.

3)

In readwriteval.F90 I have re-ordered some declarations.  When declaring something like

integer :: a(b,c)

the b and c have to be declared before a; the rearrangements make sure this is the case.

4)

For the rest the exit function is not standard Fortran and for NAG is accessed via a compiler specific module.  I have done this hidden behind a CPP flag.

### DIFF
--- a/utils/apxdx.F90
+++ b/utils/apxdx.F90
@@ -21,6 +21,10 @@
 !> @author J. Ator @date 2009-07-01
 program apxdx
 
+#ifdef NAGFOR
+  use f90_unix_proc, only: exit
+#endif
+
   implicit none
 
   character*240 cdxtbl, cbffil
--- a/src/mastertable.F90
+++ b/src/mastertable.F90
@@ -566,7 +566,7 @@ subroutine sntbbe ( ifxyn, line, mxmtbb, nmtbb, imfxyn, cmscl, cmsref, cmbw, cmu
   implicit none
 
   integer, intent(in) :: ifxyn, mxmtbb
-  integer, intent(out) :: nmtbb, imfxyn(*)
+  integer, intent(inout) :: nmtbb, imfxyn(*)
   integer ntag, ii, nemock
 
   character, intent(out) :: cmelem(120,*), cmunit(24,*), cmsref(12,*), cmmnem(8,*), cmscl(4,*), cmbw(4,*), cmdsc(*)*4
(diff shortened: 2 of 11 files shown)


edwardhartnett:
Is there a way we can download the NAG compiler and use it in GitHub?

Otherwise, how can we know that the NAG build does not break in the future?

DJDavies2:
I don't know about NAG and github.  Beyond that the only solution would be to wait for people who are testing with it to raise PRs.

edwardhartnett:
Seems like NAG is the one compiler that cannot run on GitHub due to weird license issues.

I wonder if any of these could be detected as GNU warnings. See #300.

utils/apxdx.F90
```diff
@@ -21,6 +21,10 @@
 !> @author J. Ator @date 2009-07-01
 program apxdx
 
+#ifdef NAGFOR
```
edwardhartnett, line 24: We really don't want compiler specific ifdefs in the code. Is there another way we can resolve this?

What is the problem with exit?

DJDavies2: The problem is that exit is not a standard Fortran intrinsic, so compilers don't have to provide it.

DJDavies2: Maybe STOP and/or ERROR STOP would be a standardly correct alternative?  They can return error codes to the environment.

DJDavies2:
> Seems like NAG is the one compiler that cannot run on GitHub due to weird license issues.
> 
> I wonder if any of these could be detected as GNU warnings. See #300.

The exit statement might be detectably by gnu if there is a warning to detect non-standard intrinsics.  Not sure about the declaration order stuff.  The intent(out) would need runtime checks I would guess, and then it would depend on how good gfortran's undefined variable checking is.

jbathegit:
@DJDavies2 thanks very much for reporting this, as I'd never even heard of the NAG compiler until now.  But we definitely want our codes to be as portable as possible, so this is extremely helpful!

Even though there's no NAG compiler for testing in GitHub, your suggestions nevertheless worked without any problems on all of the other compilers we test with, so I've gone ahead and added them in as part of a separate PR #638.  For the exit function issue, I agree with @edwardhartnett that we want to stay away from compiler-specific preprocessor directives, so I've just gone ahead and changed all of those to equivalent stop statements.

I think PR #638 now covers everything you've noted here, but this is your issue so I'll leave it to you to close it if you agree.  If not, let us know.

Thanks again!

jbathegit:
> I think PR #638 now covers everything you've noted here, but this is your issue so I'll leave it to you to close it if you agree. If not, let us know.

We haven't heard anything further, so I'm going to go ahead and close this as an abandoned PR.  If anything changes, please feel free to reopen or start a new one.

PR #647, opened 2025-03-30
jsw-sinv-ufbtab

### LINKED ISSUE #579: Limit for sinv needs to be increased
When running sinv  on a large tank, I get an error message.

[clogin07 /lfs/h2/emc/obsproc/noscrub/sudhir.nadiga]$ ls -l /lfs/h1/ops/para/dcom/20240318/b021/xx054
-rw-rw-r-- 1 ops.para para 2073922960 Mar 19 01:16 /lfs/h1/ops/para/dcom/20240318/b021/xx054
[clogin07 /lfs/h2/emc/obsproc/noscrub/sudhir.nadiga]$ 

**ERROR MESSAGE BELOW**
[clogin07 /lfs/h2/emc/obsproc/noscrub/sudhir.nadiga]$ sinv /lfs/h1/ops/para/dcom/20240318/b021/xx054
 +++++++++++++++++++++WARNING+++++++++++++++++++++++
 BUFRLIB: UFBTAB - THE NO. OF DATA SUBSETS IN THE BUFR FILE IS .GT. LIMIT OF 16000000 IN THE 4TH ARG. (INPUT) - INCOMPLETE READ  
 >>>UFBTAB STORED 15999291 REPORTS OUT OF ********<<<                                                                            
 +++++++++++++++++++++WARNING+++++++++++++++++++++++
  
 +++++++++++++++++++++WARNING+++++++++++++++++++++++
 BUFRLIB: UFBTAB - THE NO. OF DATA SUBSETS IN THE BUFR FILE IS .GT. LIMIT OF 16000000 IN THE 4TH ARG. (INPUT) - INCOMPLETE READ  
 >>>UFBTAB STORED 15999291 REPORTS OUT OF ********<<<                                                                            
 +++++++++++++++++++++WARNING+++++++++++++++++++++++
  
 
209  NOAA 18              9109137  000                                                                                
223  NOAA 19              6890154  000                                                                                

                         15999291

[clogin07 /lfs/h2/emc/obsproc/noscrub/sudhir.nadiga]$ 

How do we address this issue? Thanks.

jack-woollen:
Fixes #579

Subroutine UFBTAB normally will attempt to read through all BUFR messages in a file to accumulate a table of elements specified in the argument list into an out array. If the number of messages in the file exceeds to amount of array space, it will stop storing values, compile a message count, and then return with an informative message. This is inconvenient for making inventories of elements in large BUFR files, so a "part" switch is implemented (in subroutine SETPART) to allow an app to repeatedly call UFBTAB and receive output for the entire file in parts (or chunks) until all requested data is read and returned. The builtin utility "sinv" is an example of an app which benefits from the expansion of UFBTAB capability.


jbathegit:
Thanks again for this Jack, but I just checked the gcovr output and it looks like we are going to need to add a test case for the new ufbtab() lines 917-923 and 986-987, b/c they aren't being hit by any of the existing sinv test cases.  And now that I think of it that makes sense, b/c none of the input files for any of the existing sinv tests have anywhere close to the 16000000 data subset limit that we're using to dimension the said array, and so those new lines of ufbtab() are never being hit.

So we're going to need to add a test for those lines, and I can work to do that and then push up a follow-on commit on top of what you've submitted so far for this PR.  But before I do that I need to think carefully about the best way to do that, b/c it won't be ideal to add a monster file the size of, say, one of the existing b012/xx024 tanks to our existing testfiles tarball just to test this new chunking logic.  So I'm wondering if it might make sense and help the cause to maybe now lower the value of maxa from 16000000 to something a bit smaller, and now that we're going to be adding this new chunking logic into the sinv utility to allow it to handle a satellite file of any size.  Thoughts?  Or would doing that somehow negatively impact the performance of sinv, by forcing it to now have to use the chunking logic more often than it would otherwise if we had left maxa unchanged?

jack-woollen:
I was thinking that the most economical way to test the chunking logic is just pass a smaller value for i2 into ufbtab for one (or more) of the tests. No need even to have a bigger or smaller array in the test program. But either way, anything that lowers i2 below the amount of data in the file, after setting part true, tests the chunking.

jbathegit:
OK that makes sense, and I'll give it a try and keep you posted - thanks!

jbathegit:
Hey Jack, just to give you an update, I've been working on this in between other tasks, and I'm close to having a new commit ready to push up for you to look at.  I actually had to make some additional tweaks to ufbtab, in addition to the new test program. 
 But I've just about run out of steam for today, and I'm off tomorrow, so it will likely be Wednesday morning at the earliest before I can push my changes up for you to review.  Again, I'll keep you posted, but just didn't want you to worry that I'd forgotten about this :-)

jbathegit:
Hey Jack, I just pushed up a new commit on top of your earlier ones.  I think we're good to go now, but please take a look at my updates and let me know if you have any questions or concerns before we merge it.

As I noted earlier, and in addition to the new intest13.F90 test code, I also had to make some tweaks to ufbtab.  These showed up as I was testing the new code lines 917-923 and 986-987, and while using the developer-level enhanced compiler flags where optimization is temporarily disabled.  For example, during successive calls to ufbtab (i.e. when the "goto 11" jump was happening), the just_count variable was now unset, and which in turn led to unpredictable behavior in the "outer do while" loop since just_count is the controlling variable there.  Similarly, the openit variable was also now unset during successive calls, which in turn would lead to unexpected behavior such as trying to call rewnbf to try to restore lunit to its earlier settings even though it had never been rewound in the first place, and therefore leading to a bort exit.  And the lun variable also needed to be explicitly saved during the initial call to ufbtab, since subroutine status was no longer being called during later successive calls to ufbtab, and therefore the value of lun was now also unset during those successive calls and leading to unpredictable behavior when trying to call string, ustrpl, etc.

No worries, but again, a lot of these sorts of things only really jump out when you've got all of the enhanced debug and warning flags turned on and optimization turned off, and otherwise they can be very easy to miss.  I also condensed the documentation a bit for consistency with the existing documentation, including the use of parentheses after subroutine names to auto-generate links within Doxygen, and to also fully describe the iret parameter within the ufbtab docblock, since it's now both an input and output parameter to that routine.

Again, thanks for all of your help with this task, and please let me know if you have any additional questions or concerns!

jbathegit:
Fixes #579

jbathegit (approved):

ilianagenkova:
After looking at "Files changed" I want to confirm,  should we use _setpart ( **x**part )_ or _setpart ( part )_  (to turn the feature on) prior to the UFBTAB calls in applications like bufr-dump routines?

Also, this functionality would require we use the latest bufr.v12.\*.\* version - what version would that be exactly? Thanks.

jbathegit:
> After looking at "Files changed" I want to confirm, should we use _setpart ( **x**part )_ or _setpart ( part )_ (to turn the feature on) prior to the UFBTAB calls in applications like bufr-dump routines?

You would just use `call setpart(.true.)`.  You can see an example of how to do this in the new test/intest13.F90 code.
> 
> Also, this functionality would require we use the latest bufr.v12.*.* version - what version would that be exactly? Thanks.

TBD at this point, b/c we don't have a date yet for the next release, nor do we know what exactly that version number will be.  In the meantime, if you want to test it out, you can just clone the latest version of the develop branch from GitHub.  After you do that, just follow the README instructions to make your own build of that clone, and then the new version of the sinv utility will be in the resulting build/utils directory for you to test with.

PR #649, opened 2025-04-07
doxygen updates

AlexanderRichert-NOAA:
This PR updates the Doxygen to include a table of CMake options, and makes a few other updates to facilitate compatibility with more recent Doxygen versions. It also adds a custom CSS stylesheet so that `inline code snippets` in Doxygen look nice when rendered into HTML.

I rewrote the installation instructions to be consisent with other NCEPLIBS and to be consistent with what's recommended by CMake developers, but I'm not married to it if you guys prefer the existing way.


jbathegit:
Thanks for this Alex!  I just made a couple of minor tweaks, the main one being that I don't believe the build directory should ever be created inside of the main NCEPLIBS-bufr directory.  At least I've always done it elsewhere to keep it outside of the actual package that gets cloned from GitHub, and so that it never accidentally gets pushed back up to the server after, say, a `git commit -a`.   My $0.02 anyway :-)

jbathegit (approved):

jbathegit:
Hey Alex, after thinking about this a bit more, I realized that most folks who will end up following these instructions are just downloading and building the software for their own use, versus someone like me who may also at some point be pushing software updates back up to the server.  So what you had initially makes a lot more sense for the majority of folks, to just build everything in a new build subdirectory within the actual package to keep everything together, versus in a separate, outside directory location like I normally do.

So bottom line, I'll go ahead and push up a follow-on commit as a new PR to restore those particular details to what you had earlier :-)

PR #654, opened 2025-04-28
Add ncepbufr._bufrlib.writlc and write_long_string to Python API.

DanielO:
I needed this to put PTIDC into a file - it has only had very light testing.

### DIFF
--- a/python/_bufrlib.pyf
+++ b/python/_bufrlib.pyf
@@ -148,6 +148,11 @@ end subroutine writsb
 subroutine cmpmsg(cmp) ! in cmpmsg.f
     character*(*),intent(in) :: cmp
 end subroutine cmpmsg
+subroutine writlc(lunit,chr,st) ! in readwriteval.F90
+    integer,intent(in) :: lunit
+    character*(*),intent(in) :: chr
+    character*(*),intent(in) :: st
+end subroutine writelc
 
     end interface 
 end python module _bufrlib
--- a/python/ncepbufr/__init__.py
+++ b/python/ncepbufr/__init__.py
@@ -446,6 +446,27 @@ class open:
         except Exception as error:
             print(f"An exception occurred {error}")
         return result
+    def write_long_string(self,s,mnemonic,first=False):
+        """
+        Encode long character string and write to the current message.
+        String must be ASCII encodable otherwise an exception will be
+        raised.
+
+        If first is true then writsb will be called first.
+
+        Example:
+
+            :::python
+            >>> bufr = ncepbufr.open(filename)
+            >>> bufr.write_long_string('test123',mnemonic='PTIDC',end=True)
+        """
+        if len(mnemonic.split()) > 1:
+            raise ValueError('only one mnemonic per call to write_long_string')
+        if first:
+            _bufrlib.writsb(self.lunit)
+        dat = s.encode('ASCII')
+        _bufrlib.writlc(self.lunit,dat,mnemonic)
+
     def read_subset(self,mnemonics,rep=False,seq=False,events=False):
         """
         decode the data from the currently loaded message subset


jbathegit:
@DanielO thanks for submitting this, but I would be curious to know the results of your testing, because I think you may have a logic error in your proposed new write_long_string routine, and so I don't see how you would ever obtain the results you expect. 

Specifically, before you call the library subroutine writlc, you need to first call writsb.  This is noted in the [documentation for writlc](https://noaa-emc.github.io/NCEPLIBS-bufr/readwriteval_8F90.html#aecfa08f631e1af46b472c007b93b2955), and I realize it's counterintuitive to have to close out a subset using writsb before you write out a long string into that same subset using writlc, but this is due to the peculiar underlying way that writlc works, and which admittedly is different from all of the other routines which write data values into a subset.

So bottom line, I think you need to refactor your logic to call _bufrlib.writsb(self.lunit) directly within your Python application code, or else indirectly via the existing write_subset routine using end=.True.  But I really don't think you should ever call it from within your new write_long_string routine, because it will always need to be done before you can make **_any_** call to write_long_string!  So the only way you could make it work inside of your new write_long_string routine is if you swapped out your new end=.True. boolean for a first=.True. boolean instead, because you'd only ever want to call _bufrlib.writsb(self.lunit) internally during the first call to write_long_string by the application program, and you'd never want to call it again if you had additional long character strings to store via additional subsequent calls to write_long_string within your application program!  And of course you'd also have to call _bufrlib.writsb(self.lunit) **_before_** (vs. after) your internal call to _bufrlib.writlc(self.lunit,dat,mnemonic), as noted above.

Hopefully I've been able to explain that clearly, but in any case I still think that's very kludgey, and so again I would recommend that any Python application always be required to just explicitly close out a subset on its own via a direct call to _bufrlib.writsb(self.lunit) (or else via the existing write_subset routine using end=.True.), and before making any subsequent calls to your new write_long_string routine.

DanielO:
I added the end argument to be like the other calls but ended up not using it because it produced some odd results (for the reasons you suggest I imagine) so I guess it should be removed.

I call it like so:
```
import ncepbufr
h = ncepbufr.open('test.bufr', mode = 'w', table = 'test.table')
h.open_message('TEST', 0)
hdstr = 'WGOSIDS'
hdr = h.missing_value * numpy.ones(len(hdstr.split()), float)
hdr[0] = 0
h.write_subset(hdr, hdstr, end = True)
h.write_long_string('test123', 'LSTN')
h.close_message()
h.close()
```

And test.table looks like so:
```
|------------------------------------------------------------------------------|
| MNEMONIC | NUMBER | DESCRIPTION                                              |
| TEST     | A48122 | TEST HEADER                                              |
| WGOSIDS  | 001125 | WIGOS identifier series                                  |
| LSTN     | 001019 | Long station or site name                                |
|------------------------------------------------------------------------------|
| MNEMONIC | SEQUENCE                                                          |
| TEST     | WGOSIDS LSTN
|------------------------------------------------------------------------------|
| MNEMONIC | SCAL | REFERENCE   | BIT | UNITS                    |-------------|
| WGOSIDS  |    0 |           0 |   4 | Numeric                  |-------------|
| LSTN     |    0 |           0 | 256 | CCITT IA5                |-------------|
```

DanielO:
I've changed it to first as you suggest and this now works:
```
import ncepbufr
h = ncepbufr.open('test.bufr', mode = 'w', table = 'test.table')
h.open_message('TEST', 0)
h.write_long_string('test123', 'LSTN', first = True)
h.close_message()
h.close()
```

jbathegit (approved):

jbathegit:
Thanks again @DanielO for submitting and sharing this with us!

PR #657, opened 2025-05-04
Use setjmp/longjmp to convert bort/bort2 errors into python exceptions.

DanielO:
Replaces borts.F90 with borts.c that has hooks to call longjmp on an error.
Update the Python API wrapper to use these hooks to convert bort/bort2 calls into python exceptions.

I will be the first to say this is pretty kludgy, however having the library call exit when you make a coding mistake makes it annoying to use and also makes it more risky to use in a long running application (like a data processing deemon or Jupyter notebook etc)

Sample output:
```
>>> import ncepbufr
>>> ncepbufr._bufrlib.openbf(1, 'a', 1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
_bufrlib.NCEPBufrError: BUFRLIB: OPENBF - ILLEGAL SECOND (INPUT) ARGUMENT
>>>
```


DanielO:
I tried a few things to fix the undefined behaviour in the test but I can't work out what the underlying problem is.
So far as I can see the string passed to puts should be fully defined with a terminating \0 (strcat guarantees that).
I built a local version with -fsanitise=address but I was not able to reproduce the error (using macOS with GCC14 built by macports)

DanielO:
I appear to have mangled this bringing it up to date, will try again.

PR #660, opened 2025-05-29
more Python updates

### LINKED ISSUE #375: WAY too much printed to stdout during tests
Trying to find a failing test among ~15K lines of output is impossible.

What tests are writing 10s of thousands of lines of output, and what is achieved by this?

Output from tests should be short so we can easily find which test failed in CI.

jbathegit:
Fixes #375

Also adds new Python wrapper, removes superfluous tests, enhances documentation, and improves CI verification


aerorahul (approved):
@jbathegit 
Rather than throwing away the superflous tests, what if they served as `examples/`?

jbathegit:
Good question Rahul, and the answer is that the same exact code for each of those deleted files was already present in the test_misc.py file.

Furthermore, test_misc.py also contained actual verification checks of all of the generated output, but none of those same checks were present in any of the deleted files.  So test_misc.py was more comprehensive, and the deleted files contained nothing of any additional value (and weren't even really "tests" anyway in the CI sense), which in turn made them duplicative (and therefore superfluous :-)

aerorahul:
@jbathegit 
Thanks for that explanation.  I am good with removing the files as their essence is captured in the `test_misc.py`.

Issue #662, opened 2025-06-03
Enabling compression breaks writing strings.


DanielO:
I have the following table file:
```
.------------------------------------------------------------------------------.
| ------------   USER DEFINITIONS FOR TABLE-A TABLE-D TABLE B   -------------- |
|------------------------------------------------------------------------------|
| MNEMONIC | NUMBER | DESCRIPTION                                              |
|----------|--------|----------------------------------------------------------|
|          |        |                                                          |
| TESTSTR  | A48122 | TEST                                                     |
|          |        |                                                          |
| WGOSIDS  | 001125 | WIGOS identifier series                                  |
| LSTN     | 001019 | Long station or site name                                |
|          |        |                                                          |
|------------------------------------------------------------------------------|
| MNEMONIC | SEQUENCE                                                          |
|----------|-------------------------------------------------------------------|
|          |                                                                   |
| TESTSTR  | WGOSIDS LSTN                                                      |
|          |                                                                   |
|------------------------------------------------------------------------------|
| MNEMONIC | SCAL | REFERENCE   | BIT | UNITS                    |-------------|
|----------|------|-------------|-----|--------------------------|-------------|
|          |      |             |     |                          |-------------|
| WGOSIDS  |    0 |           0 |   4 | Numeric                  |-------------|
| LSTN     |    0 |           0 | 256 | CCITT IA5                |-------------|
|          |      |             |     |                          |-------------|
`------------------------------------------------------------------------------'
```

And the following python file:
```
#!/usr/bin/env python3

import ncepbufr
import numpy

h = ncepbufr.open('teststr.bufr', mode = 'w', table = 'teststr.table')
ncepbufr.section01_write('BEN', 4)
#ncepbufr.standard_messages(True)
h.cmpmsg('Y')

hdstr1 = 'WGOSIDS'

hdr1 = h.missing_value * numpy.ones(len(hdstr1.split()), float)

hdr1[0] = 0
h.open_message('TESTSTR', 2025010100)
h.write_subset(hdr1, hdstr1, end = True)
h.write_long_string('test 123', 'LSTN')
h.close_message()
h.close()
```

It runs but debufr shows:
```
MESSAGE TYPE TESTSTR

001125  WGOSIDS                        0  NUMERIC                       WIGOS identifier series
001019  LSTN                                          ( 32)CCITT IA5                Long station or site name
```

If I comment out the `h.cmpmsg('Y')` line then it is:
```
[...]

PR #667, opened 2025-06-18
remove unneeded line from stndrd logic

### LINKED ISSUE #661: Error when stdmsg is enabled.
I have the following table file:
```
.------------------------------------------------------------------------------.
| ------------   USER DEFINITIONS FOR TABLE-A TABLE-D TABLE B   -------------- |
|------------------------------------------------------------------------------|
| MNEMONIC | NUMBER | DESCRIPTION                                              |
|----------|--------|----------------------------------------------------------|
|          |        |                                                          |
| TESTSTR  | A48122 | TEST                                                     |
|          |        |                                                          |
| WGOSIDS  | 001125 | WIGOS identifier series                                  |
| LSTN     | 001019 | Long station or site name                                |
|          |        |                                                          |
|------------------------------------------------------------------------------|
| MNEMONIC | SEQUENCE                                                          |
|----------|-------------------------------------------------------------------|
|          |                                                                   |
| TESTSTR  | WGOSIDS LSTN                                                      |
|          |                                                                   |
|------------------------------------------------------------------------------|
| MNEMONIC | SCAL | REFERENCE   | BIT | UNITS                    |-------------|
|----------|------|-------------|-----|--------------------------|-------------|
|          |      |             |     |                          |-------------|
| WGOSIDS  |    0 |           0 |   4 | Numeric                  |-------------|
| LSTN     |    0 |           0 | 256 | CCITT IA5                |-------------|
|          |      |             |     |                          |-------------|
`------------------------------------------------------------------------------'
```

And the following python file:
```
#!/usr/bin/env python3

import ncepbufr
import numpy

h = ncepbufr.open('teststr.bufr', mode = 'w', table = 'teststr.table')
ncepbufr.section01_write('BEN', 4)
ncepbufr.standard_messages(True)
#h.cmpmsg('Y')

hdstr1 = 'WGOSIDS'

hdr1 = h.missing_value * numpy.ones(len(hdstr1.split()), float)

hdr1[0] = 0
h.open_message('TESTSTR', 2025010100)
h.write_subset(hdr1, hdstr1, end = True)
h.write_long_string('test 123', 'LSTN')
h.close_message()
h.close()
```

However when I run it I get a `BUFRLIB: STNDRD - BIT MISMATCH COPYING SECTION 4 FROM INPUT TO OUTPUT (STANDARD) MESSAGE` error.
Commenting out the `ncepbufr.standard_messages(True)` means it runs to completion.

jbathegit:
Fixes #661

As noted in the discussion of #661, I'm not really sure why this particular line of code is even included in the logic for subroutine stndrd.  That line of code doesn't make any sense to me, and it's never been triggered in any of our existing test cases.

It did get triggered in the example code presented in #661, and which in turn caused an abort, but then when I went and removed that line of code altogether, the example code worked.  So again, I don't believe there's any need for that particular line of code (and especially b/c I can't figure out why it's even there in the first place), and which is why I'm proposing to just remove it from the library.

@jack-woollen can you think of anything I'm missing here?  Note that if someone down the road ever shows us a specific case where that particular line of code (or maybe something similar or more intuitive?) is really necessary, then we can always restore it or otherwise update the library again at that point.  But we've now seen an example case where that line of code was actually causing a problem, so again I think we should just remove it at this point.

### DIFF
--- a/src/readwriteval.F90
+++ b/src/readwriteval.F90
@@ -288,7 +288,7 @@ recursive subroutine writlc(lunit,chr,str)
           catx(n,ncol)=' '
           ! The following statement enforces a limit of mxlcc characters per long character string when writing
           ! compressed messages. This limit keeps the array catx to a reasonable dimensioned size.
-          nchr=min(mxlcc,ibt(node)/8)
+          nchr=min(mxlcc,len(chr),ibt(node)/8)
           catx(n,ncol)=chr(1:nchr)
           call usrtpl(lun,1,1)
           return
--- a/src/standard.F90
+++ b/src/standard.F90
@@ -227,9 +227,8 @@ recursive subroutine stndrd(lunit,msgin,lmsgot,msgot)
         islen = lsub-2
       else
         ! This is the only subset in the message, and it could possibly be an overlarge (> 65530 bytes) subset, in
-        ! which case we can't rely on the value stored in the byte counter.  either way, we don't really need it.
+        ! which case we can't rely on the value stored in the byte counter.  Either way, we don't really need it.
         islen = iad4+len4-(ibit/8)
-        if (mod(len4,2)==0) islen = islen - 1
       endif
       do l=1,islen
         call upb(nval,8,msgin,ibit)
--- a/test/test_misc.F90
+++ b/test/test_misc.F90
@@ -13,10 +13,12 @@ program test_misc
   integer imsg, idate, iunit
   integer*4 ireadmm, igetfxy, isetprm
   integer mbay(2)
-  character*8 subset
+  character*8 subset, lstr
+  character*9 ptidc
   integer iupb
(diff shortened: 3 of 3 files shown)


jbathegit:
I updated test_misc.F90 case for the change noted in #661.

Also, while digging into #662, I came across a bug in subroutine writlc.  Specifically, and when compression is turned on, if the user passes in a chr string whose allocated length within the application program is less than the prescribed Table B width for the associated str mnemonic (as calculated internally within the local variable nchr), then the expression `chr(1:nchr)` ends up referencing unallocated memory and causing a SIGSEV at runtime.  The simple solution is to never attempt to store more characters than `len(chr)`, and that fixes the problem.  So I've also now added another test case to test_misc.F90 for that as well :-)

jack-woollen (approved):

PR #668, opened 2025-07-01
add bounds checking to developer CI test

### LINKED ISSUE #666: enable bounds checking in developer.yaml CI test
As a follow-on to the discussion in #664, we could enable bounds checking in the developer.yaml CI test by adding `-fcheck=all` to the CMAKE_Fortran_FLAGS in the build step.  This would obviously significantly increase the run time for the developer.yaml CI test, but the benefit is that it would prevent any further problematic code from ever being introduced into the library in the future.

However, before we can do that, we need to clean up all of the existing problematic code in the library which will currently trigger a ctest error whenever `-fcheck=all` is included in CMAKE_Fortran_FLAGS.  And there are quite a few of those! :-)

As a side note, `-fcheck=all` isn't a valid corresponding option to the GNU C compiler, so we won't be able to add that same option to the CMAKE_C_FLAGS in the same build step.

jbathegit:
Fixes #666

### DIFF
--- a/src/bufr_c2f_interface.F90
+++ b/src/bufr_c2f_interface.F90
@@ -51,13 +51,13 @@ module bufr_c2f_interface
     !>
     !> @author Ronald McLaren @date 2020-07-29
     function c_f_string(c_str) result(f_str)
-      character(kind=c_char,len=1), intent(in) :: c_str(*)
+      character(kind=c_char, len=1), intent(in) :: c_str(*)
       character(len=:), allocatable :: f_str
       integer :: nchars
 
       nchars = 1
       do while (c_str(nchars) /= c_null_char)
-          nchars = nchars + 1
+        nchars = nchars + 1
       end do
       nchars = nchars - 1
 
@@ -94,7 +94,7 @@ module bufr_c2f_interface
     !> @author Ronald McLaren @date 2020-07-29
     subroutine open_c(lunit, filepath) bind(C, name='open_f')
       integer(c_int), value, intent(in) :: lunit
-      character(kind=c_char, len=1) :: filepath
+      character(kind=c_char) :: filepath
 
       open(lunit, file=c_f_string(filepath))
     end subroutine open_c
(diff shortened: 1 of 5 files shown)


src/bufr_c2f_interface.F90
```diff
@@ -74,14 +74,15 @@ end function c_f_string
...
-      integer :: max_str_len
+      integer :: ii
 
       if (c_str_len /= 0) then
-        max_str_len = c_str_len
-        c_str(1)(1:max_str_len) = f_str(1:max_str_len)
-        c_str(1)(max_str_len:max_str_len) = c_null_char
+        do ii = 1, c_str_len
+          c_str(ii) = f_str(ii:ii)
+        enddo
```
AlexanderRichert-NOAA, line 84: Is the last character supposed to get overwritten? In any case this seems to be consistent with the existing code.

jbathegit: Yeah, now that you mention it we probably could change the top limit of that do loop to be `c_str_len - 1` instead of just `c_str_len`.  I can try doing that and pushing up another commit if you think it's worthwhile for the long term, though as you noted it is consistent with what was there before.

jbathegit: On second thought, we wouldn't want to just hardcode change the top limit to be `c_str_len - 1`, b/c if the value passed in for c_str_len was 1 then we'd end up with a top limit of the range being 0, which is less than the lower limit.

src/cidecode.F90
```diff
@@ -294,7 +294,8 @@ recursive integer function iupm(cbay,nbits) result(iret)
       ', NBITS (",I4,"), IS > THE INTEGER WORD LENGTH ON THIS MACHINE, NBITW (",I3,")")') nbits,nbitw
     call bort(bort_str)
   endif
-  cint = cbay
+  lcbay = len(cbay)
```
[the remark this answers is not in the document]
AlexanderRichert-NOAA: I ran the Copilot review here for fun, you can use this if you like but I'll resolve the conversations.

AlexanderRichert-NOAA (approved):

src/bufr_c2f_interface.F90
jbathegit: In all of the cases I can see throughout the library where this routine is being called, c_str is always explicitly allocated one larger than the maximum length of any plausible corresponding f_str.  In other words, an extra byte has already been explicitly allocated within c_str for the trailing null character that's required in C, and the value of c_str_len also already reflects that extra byte.  So at least for now, we'd never be passing in a c_str_len value of 1 unless the corresponding f_str was empty (length=0, and in which case we'd presumably never even need to call the routine in the first place :-), and that's why the logic works as it is.

PR #671, opened 2025-07-18
Use setjmp/longjmp to convert bort/bort2 errors into python exceptions.

DanielO:


jbathegit:
Hi (again :-) Daniel!  As I noted in other previous issue discussions, I do appreciate you taking the initiative to try and come up with a better and more flexible library-wide method of dealing with these sorts of errors, and if we can get past the remaining hurdles then we can certainly consider incorporating your work into the baseline develop branch.

In fact, and FYSA, our support team has already had a number of similar discussions over the years to try and come up with some minimally-impactive way to address this issue throughout the library.  See #340 for one of the more recent examples.  So if nothing else, please understand that we're already well aware of this shortcoming and would definitely like to improve on it at some point if we can (and resource permitting :-)

To that end, any such updates would still need to pass the existing developer CI test, which unfortunately is still failing in this PR.  And for any final solution, we'd also need some corresponding new test cases to add to the `test` subdirectory of the repository, and which would demonstrate the use of this new setjmp/longjmp option within at least one Fortran and one C application code, rather than just for Python.

Along those lines, I was a bit surprised by your approach in your update to `python/_bufrlib.pyf`, b/c I wasn't aware that it was even possible to embed multiple instances of a C usercode block within a Fortran interface.  In other words, I wasn't aware that you could mix C and Fortran that way within a single source file!?

DanielO:
I had another look and realised what the issue was - the Fortran strings are not necessarily nil terminated so we have to use the length. Now it passes :)

Note that there isn't a mix of C and Fortran - f2py generates C wrappers which link to Fortran so it works (albeit rather strange looking!)

The approach I used seems quite kludgy but I couldn't think of another way to do it without rewriting everything. Given the changes are fairly limited I do not think it would make it more difficult to progress #340

jbathegit:
Thanks again for your efforts on this Daniel, and this is definitely a step in the right direction!  At this point I'm adding my colleague Jack Woollen (@jack-woollen) into the conversation, since he's the other main long-time developer of the library, and he and I have both been involved in many discussions with others over the years about adding in some sort of ability for the library to gracefully pass control back to a calling program whenever a bort condition occurs, and rather than terminating the entire application from inside of the library.

So again this is good progress, but there's still some further work that we'll collectively need to complete before we could merge any of this:

1. As I noted earlier, we'd need to develop some new test cases to add to the `test` subdirectory of the repository and which utilize the new added enablesetjmp feature, from both Fortran/C and from Python.  These would also be good examples showing others how to enable this new option, which we could then also include within the [main library documentation](https://noaa-emc.github.io/NCEPLIBS-bufr/).  Could you please provide some initial examples which show how you intend for an application to utilize this new functionality, b/c I'm still fuzzy about how this would actually work in practice, and especially given that often times a bort condition can occur several nested subroutine calls down from within whatever library subroutine was directly called by the application itself.
2. Please correct me if I'm wrong, but it looks like your proposed new borts.c code is utilizing the old-style way of calling C from Fortran, including the additional implicit underscore at the end of the C function name and the implicit passing of the Fortran string length as an additional hidden call argument.  Instead, we've put a lot of effort over the past few years to modernize the C<->Fortran interface within the library to use explicit interfaces and the iso_c_binding protocol.  For examples of this, please see https://noaa-emc.github.io/NCEPLIBS-bufr/bufr__interface_8F90.html and https://noaa-emc.github.io/NCEPLIBS-bufr/bufr__c2f__interface_8F90.html, both of which contain links where you can easily view the underlying source code.  In short, I think we need to follow this same approach for transitioning from borts.F90 to borts.c, basically by documenting that interface in the source code using the bind() attribute to explicitly account for any extra Fortran string length arguments, rather than the old-style way of assuming them implicitly and via the additional underscore on the C function name.
3. On a much more minor note, and in looking at the output that's now being generated from the existing CI test cases, it looks like we also need to add a `putc('\n', stdout)` line after each of the for loops within the new bort and bort2 C functions, b/c right now the closing `***********BUFR ARCHIVE LIBRARY ABORT**************` message is being printed on the same line as the actual bort message :-)

DanielO:
Hi Jeff,
I've added a very basic Python test.

I didn't even know about the iso_c_binding so yep it is the old way :)
I tried doing it the new way but am not having much luck.
I renamed borts to borts_c and changed it to accept a C string and then created borts.F90 with:
```
subroutine bort(str) bind(C,name='bort')
  use iso_c_binding
  character*(*), intent(in) :: str
  call bort_c(f_c_string(str))
end subroutine bort
```
and update cmake to build it, but I get:
```
[ 17%] Linking Fortran executable apxdx
-macosx_version_min has been renamed to -macos_version_min
ld: warning: -ld_classic is deprecated and will be removed in a future release
Undefined symbols for architecture x86_64:
  "_bort_", referenced from:
      _status_ in libbufr_4.a(openclosebf.F90.o)
      _wtstat_ in libbufr_4.a(openclosebf.F90.o)
      _rewnbf_ in libbufr_4.a(openclosebf.F90.o)
      _posapx_ in libbufr_4.a(openclosebf.F90.o)
      _openbf_ in libbufr_4.a(openclosebf.F90.o)
      _ufbcnt_ in libbufr_4.a(openclosebf.F90.o)
      _ufbtab_ in libbufr_4.a(openclosebf.F90.o)
      ...
     (maybe you meant: _bort_c, _bort_f )
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
```
I don't understand why they are linking to `_bort_`..
Any ideas?

jbathegit:
Thanks for the basic test Daniel, but I don't see where you've activated it anywhere yet in CMakeLists.txt so that it actually gets executed for CI testing as part of this or any future PR.  So how do we know it actually works within any or all of the existing GitHub runners?

I can work to fix that and expand the testing further at some point, and we'll also need to figure out how to similarly test for abort cases which occur not only in subroutines directly called by an application, but also in any of the numerous other subroutines that may not get called directly but rather get called internally within the library, and in some cases as many as 5-6 nested levels down from inside of a direct user-called subroutine.  For example, if we now gracefully catch a jump hook signal from one of those lower-level internal subroutines, then how/where does it actually return control back to the user's application program?  Do we somehow still need to pass that signal all the way back up through the call stack, in order to hand it back to the application program from the same top-level subroutine that the application had initially directly called?

When I have some time I can also take a closer look at that `_bort_` diagnostic you mentioned.  I'm sure there's a way we can get that to work using the newer iso_c_binding protocol, but it may take a bit more digging to figure out.

At any rate, and again, I sincerely appreciate your efforts and suggestions in helping us to resolve what has (unfortunately) been a long-running overall design flaw within our library software!

jbathegit:
@DanielO and @jack-woollen, after quite a bit of trial-and-error, I've just pushed up a separate new draft PR #673 which I think will be a workable long-term solution for this issue.  It's still a draft at this point, but the main advantage in #673 (vs. here in #671 where the updates are only focused on the Python layer) is that it embeds the bort catching logic directly in the main Fortran and C code and so can also be activated for Fortran and C applications, as opposed to just for Python applications.

Please take a look and let me know what you think - thanks!

DanielO:
Hi Jeff,
#673 looks interesting but it doesn't [yet] have any Python hooks - are you going to add them or should it be done separately?

jbathegit:
> Hi Jeff, #673 looks interesting but it doesn't [yet] have any Python hooks - are you going to add them or should it be done separately?

Yes, I'll plan to include them at some point as part of that PR.  I just haven't gotten that far [yet] :-)

PR #676, opened 2025-10-10
more bort catchers

jbathegit:
Part of #340 

@jack-woollen this PR also addresses the ufbrw bug you reported via email, and note the same problem also exists in ufbrp and ufbsp and so has now been patched in those routines as well :-)

### DIFF
--- a/src/borts.F90
+++ b/src/borts.F90
@@ -108,8 +108,8 @@ end subroutine bort2
 !> @param cbc - Flag indicating whether subsequent bort errors within the NCEPLIBS-bufr
 !> software should be caught and made available to the application program via
 !> subroutine check_for_bort():
-!>  - 'N' = No (the default)
-!>  - 'Y' = Yes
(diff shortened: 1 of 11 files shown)


jack-woollen (changes requested):
The changes made in  ufbrw, ufbrp, and ufbsp, look messy to me. Those subroutines are complicated enough. I think the checks for iret<=i2 could be made more cleanly in the callers, ufbint, ufbrep, and ufbstp. And, the prints should also alert the number it tried to read as well as how many it actually did read. All three callees are already protected from going oob when iret>i2.

jbathegit:
Thanks for the suggestion @jack-woollen, and I've now pushed up an additional commit to include those changes.  Please let me know if this looks better to you.

jack-woollen (approved):
It looks much better to me. Thanks. Sorry my first impulse was to change the callee routines!

Issue #675, opened 2025-09-16
Python interface aborts program instead of raising an error on invalid operations


blaylockbk:
First off, thank you for providing a Python interface to NCEPLIBS-bufr. I'm still a novice at reading BUFR files and using the library, so forgive me if this is considered user error. 

While using the Python interface interactively to read BUFR files, I noticed that when an invalid command is run in Python, the program aborts entirely instead of raising a Python exception or error. 

For example, if I try to perform an action on a closed file:

```python
import ncepbufr

bufr = ncepbufr.open('my/bufr/file')
bufr.close()
bufr.print_table()
```
This prints the following:

```console
 ***********BUFR ARCHIVE LIBRARY ABORT**************
 BUFRLIB: DXDUMP - BUFR FILE IS CLOSED, IT MUST BE OPEN
 ***********BUFR ARCHIVE LIBRARY ABORT**************
  
8
```

…but instead of raising a Python error, it exits my program completely.

I was expecting something like a RuntimeError or ValueError to be raised in Python. Would it be possible for the Python interface to convert these aborts into Python exceptions?

jbathegit:
Thanks for reaching out Brian!  This is a known, long-standing issue in the library (see #340), and @DanielO recently asked the same question in #671 as he was working to generate interferometric radar data in BUFR.

In particular, Dan proposed a C-language approach which we're now working on incorporating as part of a new PR #673, but we're trying to do so within the internal Fortran/C structure of the main library code rather than the Python wrapper layer, so that way it can also work for Fortran and C applications instead of only for Python applications.

So bottom line, stay tuned and hopefully we'll have a more general solution in the coming months as part of the next release.

blaylockbk:
Oh, nice. Glad it's being worked on already. Didn't know about @DanielO efforts on this.

PR #677, opened 2025-10-17
add bort catching capability for C and Python applications

jbathegit:
Part of #340
Part of #675

This adds the needed functionality to catch bort errors from within C and Python applications, as shown within the included modifications to the `test/test_c_interface_2.c` and `python/test/test_misc.py` test codes, respectively.

The only reason I didn't go ahead and close either of the above issues as fully "Fix"ed is because there are still a number of library routines which need bort catching wrappers, and so that part of the overall task isn't complete yet.  But FWIW, this PR does include such a wrapper for subroutine readlc(), so at least that's one more we can check off the list :-)

### DIFF
--- a/python/ncepbufr/__init__.py
+++ b/python/ncepbufr/__init__.py
@@ -52,6 +52,22 @@ def standardize(std='N'):
     """
     _bufrlib.stdmsg(std)
 
+def bortcatch(cbc='N'):
+    """
+    Enable catching of any future bort error for return to application program via bortcheck.
+    `cbc`: `'Y'` for Yes, default is `'N'` for No, and in which case any bort error will
+    terminate the application program.
+    """
+    return _bufrlib.catch_borts(cbc)
+
+def bortcheck():
+    """
+    Check if a bort error occurred within the most-recently called BUFRLIB subroutine or function.
+    If so, then an error message is returned.  If not, then an empty string is returned.
+    """
+    bstr, bstr_len = _bufrlib.check_for_bort()
+    return bstr.strip().decode('utf-8')
+
 def set_Section01_value(s01_mnemonic, value):
     """
     Set a custom Section 0 or Section 1 message value for all subsequent writes to all bufr files,
--- a/python/_bufrlib.pyf
+++ b/python/_bufrlib.pyf
@@ -2,112 +2,99 @@ python module _bufrlib
     interface
 
 subroutine bvers(cverstr) ! in misc.F90
-    character*12,intent(out) :: cverstr
+    character*12, intent(out) :: cverstr
 end subroutine bvers
 subroutine closbf(lunit) ! in openclosebf.F90
-    integer :: lunit
+    integer, intent(in) :: lunit
 end subroutine closbf
 subroutine closmg(lunin) ! in readwritemg.F90
-    integer :: lunin
+    integer, intent(in) :: lunin
 end subroutine closmg
 subroutine copymg(lunin,lunot) ! in copydata.F90
-    integer :: lunin
-    integer :: lunot
+    integer, intent(in) :: lunin, lunot
 end subroutine copymg
 subroutine datelen(len_bn) ! in s013vals.F90
-    integer :: len_bn
+    integer, intent(in) :: len_bn
 end subroutine datelen
 subroutine dxdump(lunit,ldxot) ! in dumpdata.F90
-    integer :: lunit
-    integer :: ldxot
+    integer, intent(in) :: lunit, ldxot
 end subroutine dxdump
 subroutine fortran_open(filename, lunit, format, position, iret) ! in openclosebf.F90
     character*(*), intent(in) :: filename, format, position
-    integer,intent(in) :: lunit
-    integer,intent(out) :: iret
+    integer, intent(in) :: lunit
+    integer, intent(out) :: iret
 end subroutine fortran_open
 subroutine fortran_close(lunit,iret) ! in openclosebf.F90
-    integer,intent(in) :: lunit
-    integer,intent(out) :: iret
+    integer, intent(in) :: lunit
+    integer, intent(out) :: iret
 end subroutine fortran_close
 function getbmiss() ! in missing.F90
-    real*8 :: getbmiss
+    real*8, intent(out) :: getbmiss
 end function getbmiss
 function igetprm(cprmnm) ! in arallocf.F90
     character*(*), intent(in) :: cprmnm
     integer, intent(out) :: igetprm
 end function igetprm
 function ireadsb(lunit) ! in readwritesb.F90
-    integer :: lunit
-    integer :: ireadsb
+    integer, intent(in) :: lunit
+    integer, intent(out) :: ireadsb
 end function ireadsb
 function isetprm(cprmnm, ipval) ! in arallocf.F90
     character*(*), intent(in) :: cprmnm
     integer, intent(in) :: ipval
 end function isetprm
 function nmsub(lunit) ! in readwritemg.F90
-    integer :: lunit
-    integer :: nmsub
+    integer, intent(in) :: lunit
+    integer, intent(out) :: nmsub
 end function nmsub
 subroutine openbf(lunit,io,lundx) ! in openclosebf.F90
-    integer :: lunit
-    character*(*) :: io
-    integer :: lundx
+    integer, intent(in) :: lunit, lundx
+    character*(*), intent(in) :: io
 end subroutine openbf
 subroutine openmb(lunit,subset,jdate) ! in readwritemg.F90
-    integer :: lunit
-    character*(*) :: subset
-    integer :: jdate
+    integer, intent(in) :: lunit, jdate
+    character*(*), intent(in) :: subset
 end subroutine openmb
 subroutine readmg(lunxx,subset,jdate,iret) ! in readwritemg.F90
-    integer :: lunxx
+    integer, intent(in) :: lunxx
     character*8, intent(out) :: subset
-    integer,intent(out) :: jdate
-    integer,intent(out) :: iret
+    integer, intent(out) :: jdate, iret
 end subroutine readmg
 subroutine rewnbf(lunit,isr) ! in openclosebf.F90
-    integer,intent(in) :: lunit
-    integer,intent(in) :: isr
+    integer, intent(in) :: lunit
+    integer, intent(in) :: isr
 end subroutine rewnbf
 subroutine rtrcpt(lunit,iyr,imo,idy,ihr,imi,iret) ! in tankrcpt.F90
-    integer,intent(in) :: lunit
-    integer,intent(out) :: iyr,imo,idy,ihr,imi,iret
+    integer, intent(in) :: lunit
+    integer, intent(out) :: iyr, imo, idy, ihr, imi, iret
 end subroutine rtrcpt
 subroutine setbmiss(xmiss) ! in missing.F90
-    real*8 :: xmiss
+    real*8, intent(in) :: xmiss
 end subroutine setbmiss
 subroutine strcpt(cf,iyr,imo,idy,ihr,imi) ! in tankrcpt.F90
-    character*1 :: cf
-    integer :: iyr
-    integer :: imo
-    integer :: idy
-    integer :: ihr
-    integer :: imi
+    character*1, intent(in) :: cf
+    integer, intent(in) :: iyr, imo, idy, ihr, imi
 end subroutine strcpt
 subroutine ufbcpy(lubin,lubot) ! in copydata.F90
-    integer :: lubin
-    integer :: lubot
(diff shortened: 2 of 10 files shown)


python/ncepbufr/__init__.py
```diff
@@ -52,6 +52,22 @@ def standardize(std='N'):
     """
     _bufrlib.stdmsg(std)
 
+def bortcatch(cbc='N'):
```
aerorahul, line 55: Is there a reason, a `Y` or `N` was chosen over `0|1`.  Wouldn't this be prone to case sensitivity?

jbathegit: Thanks @aerorahul for taking a look at this PR!

To answer your question, the library already has an internal mechanism to capitalize any such input, so 'y' will already automatically get treated the same as 'Y' (and likewise 'n' the same as 'N').  This is noted in the docblocks for the underlying stdmsg and catch_borts library routines, as well as other library routines such as cmpmsg, codflg, strcpt, etc. whose own capabilities similarly need to be opted into via an affirmative choice by the user.  So bottom line, case sensitivity isn't an issue.

As for why the selection is via character Y/N vs integer 0/1, that's basically just how it was originally done in the oldest of the above routines, and as we added newer ones we kept to the existing practice for a consistent user interface and to avoid a mish-mash of options.

jack-woollen (commented):
I'd need to look at this much more closely before I'd feel competent to approve a lot of these changes. The rest is cosmetic and testing, which is fine. One question, how hard would it be to provide a trace-back when bort terminates a run?

jbathegit:
Thanks @jack-woollen for taking an initial look, and please take whatever time you need to review and familiarize yourself with these and all other proposed updates.  Or feel free to follow-up with questions if anything doesn't make sense.  The whole point of a code review is so that we can all collectively discuss and gain understanding of proposed library updates before they become part of the code baseline, and so that we can all continue to collectively support and maintain the library as a team.

To answer your question, I honestly have no idea if or how we could provide any sort of stack traceback to a user in a controlled way.  You may recall from some of our earlier discussions that one of the goals was to avoid having to refactor a lot of existing library routines to add return codes all the way down and back up through a call stack.  And in fact, one of the stated benefits of the setjmp/longjmp mechanism in the C language was that it allowed for non-local jumps directly back to the top of a call stack, and thereby avoided having to maintain that sort of long breadcrumb trail down and back up through the stack.

In my view, and given that the whole idea is that the user should directly call check_for_bort after each and every call to any library routine from the application, and since any bort error message clearly describes not only the error itself but also what downstream routine it occurred in, then it should be relatively straightforward to go back and figure out a traceback path afterwards if the user really wanted to do that.  But I don't believe that's something we should ever carry around ourselves inside of the code, and again I have no idea how or even if it would be possible to do that.

AlexanderRichert-NOAA:
I'll review this in a little while. Outside of this PR, @jbathegit, do you think it would be appropriate to add a bit of documentation for the users who want to do tracebacks as you're describing (in, say, gdb)?

jbathegit:
Thanks @AlexanderRichert-NOAA, but to clarify I wasn't suggesting adding any sort of automated traceback during program execution; rather, what I meant was that (1) assuming the program is calling check_for_bort after each direct call to a library routine (as it should be if it previously enabled that capability via catch_borts :-), and (2) that any bort error message definitively tells you in which same or lower-level routine it was triggered, then it should be easy after the fact for anyone to go back and manually figure out the call sequence which led from the directly-called routine down to the routine where the bort occurred, and especially since the [library Doxygen documentation](https://noaa-emc.github.io/NCEPLIBS-bufr/index.html) already provides "References" (i.e. this routine calls...) and "Referenced by" (i.e. this routine is called by...) lists for every single routine in the library.

Now that said, gdb would certainly be one way to do the same thing in a pseudo-automated way after the fact, but I don't feel like it's our responsibility to provide users with documentation on how to use gdb.  And besides, gdb may be of limited use anyway on the same program executable unless it and the library itself were both recompiled with -g turned on and with optimization turned off.

My $0.02 anyway :-)

AlexanderRichert-NOAA:
To clarify, I guess what I should have said was, _if_ these new mechanisms change the way that a developer would go about debugging in gdb, then that should be documented, but it sounds like it wouldn't, in which case I agree we don't need to provide generic documentation on using gdb.

jack-woollen:
!> @note Application programs should exercise caution when the catching and
!> returning of bort errors is enabled.  Specifically, and depending on the nature or severity
!> of any error caught or its depth within the internal call stack, there's no guarantee that
!> the library remains fully functional or in a useable state (for example, certain
!> intermediate values may not have been restored to previous settings, internal memory may not
!> have been fully deallocated, etc.).  Rather, the intent of this function is to allow an
!> application program to gracefully exit if a serious NCEPLIBS-bufr error does occur,
!> including potentially cleanly disengaging from other linked libraries or tasks.  So unless
!> the caught error is fairly benign and local to the subroutine or function in question, it
!> may not be possible for the application program to make an adjustment and then reattempt
!> another call to the library without leading to unpredictable results.  Instead, and if the
!> application program intends to continue running, the safest option in such cases may be
!> to use subroutine exitbufr() to fully reset the NCEPLIBS-bufr software before attempting
!> any future calls to the library.

How does the library defend against an infinite loop described above?

jack-woollen:
On the subject of trace backs, a number of Fortran compilers have a traceback or backtrace function, including intel, gnu, and gfortran.  One of these functions could be called, in the bort step, depending on which compiler is active, and if it has such a function. This would then produce trace backs as part of bort processing when using one of the traceback enabled compilers.

jbathegit:
> !> @note Application programs should exercise caution when the catching and !> returning of bort errors is enabled. Specifically, and depending on the nature or severity !> of any error caught or its depth within the internal call stack, there's no guarantee that !> the library remains fully functional or in a useable state (for example, certain !> intermediate values may not have been restored to previous settings, internal memory may not !> have been fully deallocated, etc.). Rather, the intent of this function is to allow an !> application program to gracefully exit if a serious NCEPLIBS-bufr error does occur, !> including potentially cleanly disengaging from other linked libraries or tasks. So unless !> the caught error is fairly benign and local to the subroutine or function in question, it !> may not be possible for the application program to make an adjustment and then reattempt !> another call to the library without leading to unpredictable results. Instead, and if the !> application program intends to continue running, the safest option in such cases may be !> to use subroutine exitbufr() to fully reset the NCEPLIBS-bufr software before attempting !> any future calls to the library.
> 
> How does the library defend against an infinite loop described above?

I don't see where I mentioned anything about an "infinite loop" here, unless you're talking about where I noted the possibility of "unpredictable results"(?)  In any case, there should be no danger of an infinite loop from the bort catching itself, as any caught error will immediately non-local jump back to the top of the stack and return immediate control to the application program.  The overall point I was trying to make in this note is that, since any non-local jump (by definition) doesn't unwind everything cleanly all the way back up the call stack, then there's no guarantee that the library remains in a clean/useable state, depending on what exactly it was doing at the time and/or how deep that call stack may have been before hitting the error.  So this note is basically a "user beware" message, or disclaimer, or whatever else you want to call it.  And this seemed reasonable, given that the main thing Ed H. and other users had been clamoring for over the years was to stop having the library directly abort the application in such cases but rather just return control back to the application with an indication of what the error was and then let the application program make that decision.  And it can now do those things, but at the same time we also don't want users to have a false sense of security that any such caught error is always easily recoverable, and without any possibility of future risk if they choose to just keep plowing ahead as though nothing ever happened!

Now having said all of that, and at least in all of the fairly simple test case examples I've cooked up so far in `test/intest14.F90` (for Fortran), `test/test_c_interface_2.c` (for C) and `python/test/test_misc.py` (for Python), the error has indeed been correctable within the application program, allowing it to proceed on without incident and as we would always like/hope to be the case.  But I didn't feel like we could give that same sort of blanket guarantee for any and all bort errors that might ever happen, and again depending on the severity of the error or how deep down in the call stack it was.  I hope that makes sense!

jbathegit:
> On the subject of trace backs, a number of Fortran compilers have a traceback or backtrace function, including intel, gnu, and gfortran. One of these functions could be called, in the bort step, depending on which compiler is active, and if it has such a function. This would then produce trace backs as part of bort processing when using one of the traceback enabled compilers.

If you're talking about a function that can be portably called inside of the library source code, as opposed to just something like `-fbacktrace`, `-traceback`, etc. which are compiler-specific and have to be enabled at compile time, then could you please provide an example of how that would work?  Sorry for any ignorance on my part, but I'm not aware of such a run-time capability.

Otherwise, if you are indeed just talking about compiler options, then please note that we already automatically enable the `-traceback` option for Intel builds and the `-fbacktrace` option for GNU builds, as you can see in the CMakeLists.txt file in the main repository directory.  Those options are already well known to not come with any additional run-time cost, and which is why we already have them automatically enabled.  But again, if you're talking about something else entirely, then please enlighten me! :-)

jack-woollen:
> > On the subject of trace backs, a number of Fortran compilers have a traceback or backtrace function, including intel, gnu, and gfortran. One of these functions could be called, in the bort step, depending on which compiler is active, and if it has such a function. This would then produce trace backs as part of bort processing when using one of the traceback enabled compilers.
> 
> If you're talking about a function that can be portably called inside of the library source code, as opposed to just something like `-fbacktrace`, `-traceback`, etc. which are compiler-specific and have to be enabled at compile time, then could you please provide an example of how that would work? Sorry for any ignorance on my part, but I'm not aware of such a run-time capability.
> 
> Otherwise, if you are indeed just talking about compiler options, then please note that we already automatically enable the `-traceback` option for Intel builds and the `-fbacktrace` option for GNU builds, as you can see in the CMakeLists.txt file in the main repository directory. Those options are already well known to not come with any additional run-time cost, and which is why we already have them automatically enabled. But again, if you're talking about something else entirely, then please enlighten me! :-)

Actually either or both should work. If the compiler option to produce a traceback is enabled and bort created some runtime error serious enough to trigger it, that would do it. Or maybe better use a fortran intrinsic if provided. Also the -g compiler option is needed. I use the intel example for my debugging when necessary. The GNU and gfortran suggestions come from google. Maybe the gfortran example works for all three as is!
Examples:
INTEL:
 subroutine subroutine_b
    call TRACEBACKQQ()) ! Print backtrace here
  end subroutine subroutine_b
GNU:
 subroutine subroutine_b
    call BACKTRACE() ! Print backtrace here
  end subroutine subroutine_b
GFORTRAN:
 subroutine subroutine_b
    bad=10./0.
  end subroutine subroutine_b

jbathegit:
Yes, the -g compiler option is already enabled as well within CMakeLists.txt for Intel and GNU builds.

jbathegit:
Just touching base.  If anyone is still actively reviewing this PR, then that's fine and please let the rest of us know, and then continue to take whatever additional time you need.

But if not, and unless there are any remaining questions or concerns, could one of you please click the approve button so we can move forward and get these updates merged into the develop branch?

aerorahul:
I am good.  Thanks for checking @jbathegit

AlexanderRichert-NOAA (approved):

jack-woollen (approved):
Sorry I had a few days of routine medical testing this week, and I didn't get back to this until now. I approve of the changes as is, but I also think there is actually no safe way to attempt remediating errors by rerunning sections where the code aborts. I think the possibility of infinite loops occurring in those situations should be noted and rerunning error prone pathways in general should be advised against. That's my 2 bits on the subject!

Also, from the traceback  discussion, I assume you agree that producing a traceback when the code is not capturing bort calls is both possible and a useful thing to build into the library.

jbathegit:
@jack-woollen, the main CMakeLists.txt already has the -g and traceback options set by default for all GNU and Intel builds, as noted earlier in this discussion thread, and regardless of whether bort errors are or aren't being caught.  So we should already be good there (i.e. it's already "built into the library" :-)

As for the caveat emptor about trying to continue on after a bort error, I agree that doing so can be dicey and unpredictable, and which is why I already included that lengthy cautionary note in the docblock for the new [catch_borts()](https://emc.ncep.noaa.gov/emc/pages/infrastructure/bufrlib_jeff/html/borts_8F90.html#a668c7c306c82fce3e5aa6b2660d76192) routine.

Now that said, in some instances the error is safe to continue on from if it's a fairly simple error that isn't nested multiple levels down the call stack, and as shown in some of the test cases inside of `test/intest14.F90`.  But yes, in general it's a risky proposition.

jack-woollen:
@jbathegit When bort is called and the program stops, no trace back is produced, because stop is not really a runtime error which triggers the automatic traceback. Something to trigger or request the trace back needs to go into the bort routine to get it. I gave examples above of some possibilities, I think the divide by zero works for all compilers. Or just a hard exit (not stop 8) after the errwrt prints. The traceback could be controlled by a logical parameter if need be. The default could be no traceback.

PR #682, opened 2025-11-13
more bort catching updates

jbathegit:
Part of #340

Note that I also modified how bort targets are internally set and unset, because this was previously being done using a Fortran boolean (.true. vs .false.), which is fine when you're trying to set targets within user-callable Fortran routines, but not so much when you're trying to do the same thing within user-callable C routines such as cobfl, crbmg, etc.  So I've now encapsulated those set/unset switches inside of new internal `bort_target_set/unset()` routines using more C-friendly (1 vs. 0, instead of .true. vs .false.) return values.  I've also gone ahead and modified all of the prior Fortran bort catchers to use this same new methodology, for consistency but also because it ends up reducing the overall number of lines of code throughout the library :-)

### DIFF
--- a/.github/workflows/MacOS.yaml
+++ b/.github/workflows/MacOS.yaml
@@ -16,8 +16,8 @@ jobs:
   MacOS:
     runs-on: macos-latest
     env:
-      FC: gfortran-12
-      F90: gfortran-12
+      FC: gfortran-13
+      F90: gfortran-13
 
     steps:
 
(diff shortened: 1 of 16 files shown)


jbathegit:
re: https://github.com/actions/runner-images/issues/12873, it looks like gfortran-12 is no longer supported for MacOS runners as of November 3rd.  So I'll try pushing up an additional update to MacOS.yaml for this PR, to now use gfortran-13 instead.

Also tagging others in @NOAA-EMC/nceplibs for their awareness.

Hang-Lei-NOAA:
@jbathegit you can try that in the workflow. Since it’s using Mac latest, gfortran-13 is current one.

jbathegit:
> @jbathegit you can try that in the workflow. Since it’s using Mac latest, gfortran-13 is current one.

Thanks Hang, and yes I did note that it was using `macos-latest`, but the yaml file also had FC and F90 explicitly set to `gfortran-12`, which I think is why it initially failed.  But it looks like it's now working with those settings changed to `gfortran-13`, so all good!

AlexanderRichert-NOAA (approved):

